PR #25, opened 2017-04-03
Next rate change

sashker:
New initial_rate and next_rate fields which upgrade dialpeer independently.


dmitry-sinina:
* Rename sql migration to 105_106.sql -  103_104 already exists at master branch.
* Squash it to 1 commit

PR #79, opened 2017-08-10
JSONAPI

Unnumbered:
private api in jsonapi format


Fivell:
@Unnumbered , @dmitry-sinina  jsonapi tells us not to use foreign keys as fields but use relationships instead, if we merge this it will be just look like json-api format but not really jsonapi.

max-konin:
@Fivell Hi. I agree with you. But it will take a long time... Because We shall have to  implement API for all model.... 
Mb remove foreign keys from attributes temporarily?

Fivell:
@max-konin in this case how do you see PATCH/POST  let's say for GatewayGroup?

Fivell:
@max-konin  I think there are 2 options 
  - implement real json-api and all take all the benefits from it.
  - do not implement json-api for private API at all.


I don't see how removing foreign keys from API helps because we do need link related objects after all.

max-konin:
@Fivell  what is problem with GatewayGroup?

max-konin:
> I don't see how removing foreign keys from API helps because we do need link related objects after all.

Then, I think we should change status to WIP and start to implement endpoint for all entities

max-konin:
@dmitry-sinina @Fivell Could you help us with error on CI. It mb related with new db's structure....

When we create some model we get error:
```
 ActiveRecord::StatementInvalid:
       PG::NotNullViolation: ERROR:  null value in column "id" violates not-null constraint
       DETAIL:  Failing row contains (null, destination_rate_policy1).
       : INSERT INTO "destination_rate_policy" ("name") VALUES ($1) RETURNING "id"
``` 

ID should be autoincremented. But I can not find it in structure.sql ....

dmitry-sinina:
@max-konin, destination_rate_policy table contains constats. It can't be modified via API or AA.

dmitry-sinina:
Same situation with:
- diversion-policies
- dump-levels
- gateway-rel100-modes
- transport-protocols
- filter-types
- rate-profit-control-modes
- sdp-c-locations
- session-refresh-methods
- dtmf-receive-modes
- dtmf-send-modes
- sensor-levels

Fivell:
@max-konin  just do not crate such records , try to select them from database instead

Fivell:
@max-konin , right

Gemfile
```diff
@@ -32,7 +32,8 @@ gem 'active_admin_sidebar', github: 'activeadmin-plugins/active_admin_sidebar'
 gem 'excelinator', github: 'livingsocial/excelinator'
 
 # REST API
-gem 'active_model_serializers', '~> 0.9.7'
+# gem 'active_model_serializers', '~> 0.9.7'
```
max-konin: remove it

dmitry-sinina:
Timezones should be read-only.

PR #83, opened 2017-08-14
Custom themes, cucumber tests

irenliber:


app/models/admin_user.rb
```diff
@@ -32,8 +32,10 @@ class AdminUser < ActiveRecord::Base
   validates_format_of :email, with: /\A([^@\s]+)@((?:[-a-z0-9]+\.)+[a-z]{2,})\z/i, unless:  proc { self.persisted? && email.blank? }
 
   after_save do
-    contact = billing_contact || build_billing_contact
-    contact.update!(email: self.email) if @email
+    unless Rails.env.test?
```
dmitry-sinina: Why did you disable it for test environment?

max-konin: @dmitry-sinina we have problem with db for test. Only "public" schema works stable

Fivell: lets wait for #85, since it is start for CI support in yeti-web

Fivell: >> we have problem with db for test. Only "public" schema works stable

@max-konin  can you bring more details please

max-konin: @Fivell All schemas except "public" do not exist in DB when specs runing. We are using PG as a docker container with this image https://hub.docker.com/r/maxkonin/yeti_pg/. Mb something wrong with configs.... But I think that all be fine after 85 PR

Fivell:
@irenliber  can you rebase on the top of the master please?

max-konin:
@Fivell @dmitry-sinina I merged this branch with master

PR #85, opened 2017-08-16
gem SecondBase

workgena:
- [x] install ``gem 'secondbase'`
- [x] fix broken task `db:structure:dump`
- [x] ActiveAdmin pages broke structure:dump. For example "ActiveAdmin::DatabaseHitDuringLoad: Your file, app/admin/routing/routing_groups.rb (line 9), caused a database error while Active Admin was loading."
- [x] `seeds.rb` to populate default application data (INSERT and pgq SELECT)
- [x] prepare test environment
- [x] rewrite Travis config
- [x] write README.md for SecondBase usage
- [x] Clean up `database.yml.*` (2 files)
- [x] Remove directories `aux` and `sql`. Fix debian-pakege and Makefile


app/models/cdr/base.rb
```diff
@@ -1,7 +1,4 @@
 class Cdr::Base < Yeti::ActiveRecord
   self.abstract_class = true
-  establish_connection "#{Rails.env}_cdr".to_sym
-
-
```
Fivell: @workgena DB_VER constant was used for performance

app/models/cdr/cdr.rb
```diff
@@ -185,8 +183,10 @@ def status_sym
   end
 
 
+  # TODO: use DisconnectInitiator#display_name instead
+  # Why in was cached this way? `DISCONNECTORS = DisconnectInitiator.all.index_by(&:id)`
   def disconnect_initiator_name
-    DISCONNECTORS[self.disconnect_initiator_id].try(:name)
```
Fivell: @workgena that was done for performance

app/models/cdr/cdr.rb
```diff
@@ -185,8 +183,10 @@ def status_sym
...
+    disconnect_initiator.try(:name)
```
Fivell: @workgena @dmitry-sinina  maybe we can use constant as we did before, I think we can make it lazy if we need, 

http://greyblake.com/blog/2014/10/05/lazy-object-pattern-in-ruby/
```
class LazyObject < ::BasicObject
  def initialize(&callable)
    @callable = callable
  end

  def __target_object__
    @__target_object__ ||= @callable.call
  end

  def method_missing(method_name, *args, &block)
    __target_object__.send(method_name, *args, &block)
  end
end


```

and then we can 

```
DISCONNECTORS = LazyObject.new { DisconnectInitiator.all.index_by(&:id) } 
```

workgena: Currently it makes only one request(from localhost):
```sh
DisconnectInitiator Load (1.5ms)  SELECT "disconnect_initiators".* FROM "disconnect_initiators" WHERE "disconnect_initiators"."id" IN (3, 1, 0)
```

app/models/cdr/cdr.rb
workgena: Done, with new commit

app/models/cdr/base.rb
workgena: Did Lazy for DB_VER, in last commit

PR #88, opened 2017-08-17
IP access API controller for iptables integration

dmitry-sinina:


Fivell:
@workgena  can you add tests?

workgena:
It would be nice to have some IP examples for AF_INET, AF_INET6 and one wrong case. Probably you done this with some specific IP.

workgena:
added tests, and rebase with current master branch

PR #92, opened 2017-08-25
validate_allow_termination

dmitry-sinina:


app/models/gateway.rb
```diff
@@ -218,13 +219,20 @@ def unlock
 
   protected
 
+  def allow_termination_can_be_enabled
+    if self.host.empty? and self.allow_termination==true
```
Fivell: @dmitry-sinina  ```self.allow_termination?``` can be used

PR #95, opened 2017-08-31
Customers API

workgena:


db/migrate/20170831101937_account_uuid.rb
```diff
@@ -0,0 +1,17 @@
+class AccountUuid < ActiveRecord::Migration
+  def up
+    execute %q{
+      CREATE EXTENSION "uuid-ossp" WITH SCHEMA public;
+
+      ALTER TABLE billing.accounts ADD uuid uuid NOT NULL DEFAULT uuid_generate_v1();
```
dmitry-sinina: uuid should be UNIQUE:
```
NOT NULL DEFAULT uuid_generate_v1() UNIQUE
```

workgena:
TODO:

- [ ] reorder migrations
- [ ] removing Contractor/Account should affect api_access (destroy/update)
- [ ] api_access for Admin API
- [ ] display Account#uuid, make it editable field in ActiveAdmin
- [ ] CDR resource

Now default api_access.allowd_ips is an ampty array, it is not '0.0.0.0/0', but works similar. Don't know what went wrong with default DB-value.

workgena:
TODO:

- [ ] move `api_access` table to schema `sys`
- [ ] remove `uuid` field from `sys.api_access`
- [ ] ActiveAdmin ApiAccess#edit page "Select an Option" can be selected as an option for accoutns. This is wrong.
- [ ]  ActiveAdmin ApiAccess#index#show pages should display "Accounts" and account#display_name divided by comma. Similar to "Routing groups" in Routing plan page. Think over n+1 requests
- [ ] ActiveAdmin Accounts#index#show should display uuid (in index page it should be last column)
- [ ] ActiveAdmin Destinations#index#show should display `id` and `uuid`(last column in index)
- [ ] fix http://127.0.0.1:3000/api/docs (at least admin-api) help: https://github.com/jejacks0n/apitome/pull/65

PR #97, opened 2017-09-04
run tests on ruby 2.3.1 and 2.3.4

dmitry-sinina:


dmitry-sinina:
because we build packages for version 2.3

PR #113, opened 2017-09-23
API Customer: CDR resource

workgena:
- [x] CDR resource for Customer API
- [x] documentation

### DIFF
--- a/app/models/cdr/cdr.rb
+++ b/app/models/cdr/cdr.rb
@@ -167,6 +167,8 @@ class Cdr::Cdr < Cdr::Base
 
   scope :account_id_eq, lambda { |account_id| where('vendor_acc_id =? OR customer_acc_id =?', account_id, account_id) }
 
+  scope :where_customer, -> (id) { where(customer_id: id) }
+  scope :where_account, -> (id) { where(customer_acc_id: id) } # OR vendor_acc_id ???
 
   scope :status_eq, lambda { |success|
     if success.is_a?(String)
--- a/app/resources/api/rest/customer/v1/cdr_resource.rb
+++ b/app/resources/api/rest/customer/v1/cdr_resource.rb
@@ -1,0 +1,64 @@
+class Api::Rest::Customer::V1::CdrResource < Api::Rest::Customer::V1::BaseResource
+  model_name 'Cdr::Cdr'
+
+  key_type :uuid
+  primary_key :uuid
+
+  attributes :time_start,
+             :time_connect,
+             :time_end,
+             :duration,
+             :success,
+             :destination_initial_interval,
+             :destination_initial_rate,
+             :destination_next_interval,
+             :destination_next_rate,
+             :destination_fee,
+             :customer_price,
+             :src_name_in,
+             :src_prefix_in,
+             :from_domain,
+             :dst_prefix_in,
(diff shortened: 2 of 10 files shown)


db/secondbase/structure.sql
```diff
@@ -375,7 +375,7 @@ CREATE TABLE cdr (
     yeti_version character varying,
     lega_user_agent character varying,
     legb_user_agent character varying,
-    uuid uuid,
+    uuid uuid DEFAULT public.uuid_generate_v1() NOT NULL,
```
dmitry-sinina: Так не нужно делать. Генерация uuid происходит в хранимке.

app/models/cdr/cdr.rb
```diff
@@ -167,6 +167,8 @@ class Cdr::Cdr < Cdr::Base
 
   scope :account_id_eq, lambda { |account_id| where('vendor_acc_id =? OR customer_acc_id =?', account_id, account_id) }
 
+  scope :where_customer, -> (id) { where(customer_id: id) }
+  scope :where_account, -> (id) { where(customer_acc_id: id) } # OR vendor_acc_id ???
```
workgena: @dmitry-sinina take a look at this query. This is ony for Customers API, so I did not add `OR vendor_acc_id = ?`, I'm not sure about this.

app/resources/api/rest/customer/v1/cdr_resource.rb
```diff
@@ -0,0 +1,56 @@
...
+  model_name 'Cdr::Cdr'
+
+  key_type :uuid
+  primary_key :uuid
+
+  attributes :time_start,
+             :time_connect,
+             :time_end,
+             :duration,
+             :success, # TODO: or it should be "status"?
```
workgena: In Admin-panel this field replaced with `status` (text). But in API i think it should be closer to real data. @dmitry-sinina what do you say?

app/resources/api/rest/customer/v1/cdr_resource.rb
```diff
@@ -0,0 +1,56 @@
...
+             :dst_prefix_in,
+             :to_domain,
+             :ruri_domain,
+             :diversion_in,
+             :local_tag,
+             :lega_disconnect_code, # TODO: or it shold be "leg_a_dc"
+             :lega_disconnect_reason, # TODO: or it should be "leg_a_reason"
+             :lega_rx_payloads,
+             :lega_tx_payloads,
+             :auth_orig_transport_protocol_id, # TODO: relation "auth_orig_transport_protocol"?
```
workgena: @dmitry-sinina perhaps it should be replaced with relation `auth_orig_transport_protocol`?

dmitry-sinina: yes

app/models/cdr/cdr.rb
dmitry-sinina: ` { where(customer_acc_id: id) }` is right.

app/resources/api/rest/customer/v1/cdr_resource.rb
dmitry-sinina: I think "success" is good for api

PR #131, opened 2017-11-05
Migrations with stop pionts

workgena:
When running several migrations at once, now it is possible to stop at any point, preventing future migrations. 

How to test it:

create several migrations
```sh
$ bundle exec rails g migration One
$ bundle exec rails g migration Two
$ bundle exec rails g migration Three
```

In second migration add method

```ruby
def stop_step
  true
end
```

Then run command `$ bundle exec rake db:migrate` and you will see

```text
"Timezone Europe/Kiev"
== 20171105085529 One: migrating ==============================================
== 20171105085529 One: migrated (0.0000s) =====================================

== 20171105085555 Two: migrating ==============================================
== 20171105085555 Two: migrated (0.0000s) =====================================

IMPORTANT: Now update and restart your servers. And after that run `rake db:migrate` again.
```

Then run `$ bundle exec rake db:migrate` again and you should see

```text
== 20171105085608 Three: migrating ============================================
== 20171105085608 Three: migrated (0.0000s) ===================================
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,3 +53,26 @@ And run tests:
 ```sh
 $ bundle exec rspec
 ```
+
+# Migrations
+
+When you run several migrations in a row, you may wish to stop at some point. In this case you should add `stop_step` method to the migration:
+
+```ruby
+# example /db/migrate/20171105085529_one.rb
+def change
+  # do something
+end
+
+def stop_step
+  true
+end
+```
+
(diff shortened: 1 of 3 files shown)


workgena:
Tests are failing. They were fixed in PR #129 We should accept #129 first, and rebase with master this branch.

PR #161, opened 2017-12-05
Batch update/delete as delayed jobs feature added. 

stekatk:
Added batch update/delete features using active_admin_scoped_collection_action gem.
Those features work as delayed jobs to handle large batches without hanging the system.
There are three Resourse DSL files:
acts_as_async_destroy.rb
acts_as_async_update.rb
acts_as_delayed_job_lock.rb

First two add batch delete and batch update features respectievly and acts_as_delayed_jobs_lock prevents changing or deleting of records of that model when there is an active batch action job to avoid some bad scenarios.

This new functionality added for the following models:
Account
Contact
Contractor
Invoice
Payment
Gateway
GatewayGroup
AreaPrefix
CustomersAuth
Destination
Dialpeer
Lnp::Cache (delete only)
Numberlist
RoutingPlan
RoutingPlanLnpRule
RoutingPlanStaticRoute


Gemfile
```diff
@@ -25,6 +25,7 @@ gem 'activeadmin', github: 'yeti-switch/active_admin'
 gem 'novus-nvd3-rails', github: 'yeti-switch/nvd3-community-rails'
 gem 'active_admin_theme'
 gem 'active_admin_import' , '3.0.0.pre'
+gem 'active_admin_scoped_collection_actions', github: 'Zamyatin-AA/active_admin_scoped_collection_actions', branch: 'batch_action_collection_fix', ref: 'd199e2867d950143a6b2c74e54c3f289d8a9b1e0'
```
workgena: This fix accepted in gem release 0.3.5. 
`gem 'active_admin_scoped_collection_actions'` should work

`$ bundle update active_admin_scoped_collection_actions` should do the trick for local update.

app/admin/billing/account.rb
```diff
@@ -4,7 +4,27 @@
   acts_as_safe_destroy
   acts_as_audit
   acts_as_clone
-
+  acts_as_async_destroy('Account')
+  acts_as_async_update('Account',
+                       lambda do
+                         {
+                         contractor_id: Contractor.all.map { |c| [c.name, c.id] },
```
workgena: I propose to simplify this `all.map {...}` everywhere to: 
```ruby
MyModel.pluck(:name, :id)
```

app/admin/billing/contractors.rb
```diff
@@ -6,10 +6,25 @@
   acts_as_clone
   acts_as_safe_destroy
   acts_as_status
+  acts_as_async_destroy('Contractor')
+  boolean = [ ['Yes', 't'], ['No', 'f']]
```
workgena: I propose(may no work) to move this variable to the method in ResourceDSL and use it instead.

```ruby
# lib/resource_dsl/acts_as_async_update.rb
module ResourceDSL
  class ActsAsAsyncUpdate
    def boolean_select_options
      [ ['Yes', 't'], ['No', 'f'] ]
    end
  end
end
```

app/jobs/async_batch_destroy_job.rb
```diff
@@ -0,0 +1,13 @@
+class AsyncBatchDestroyJob < ActiveJob::Base
+  queue_as :batch_actions
+
+  BATCH_SIZE = 1000
+
+  def perform(model_class, sql_query)
+    begin
+      scoped_records = model_class.constantize.find_by_sql(sql_query + " LIMIT #{BATCH_SIZE}")
+      scoped_records.each { |record| record.destroy! }
```
workgena: I'm no sure if it will work. If I understand this correctly the `find_by_sql ( + LIMIT 1000)` will find only first 1000 records and no more. For such cases is useful this:

http://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each

It would be nice to have one test on destroy/update more then 1000 records.

But `find_each` works only with `ActiveRecord::Relation` the original `scoped_collection_records` is an instance of ActiveRecord::Relation.

It instead of sql_query use Relation, then AsyncBatchDestroyJob and AsyncBatchUpdateJob will be match simpler.

```ruby
scoped_collection_records.find_each(batch_size: 1000) do |r|
  r.destroy!
end
```

app/jobs/async_batch_update_job.rb
```diff
@@ -0,0 +1,32 @@
+class AsyncBatchUpdateJob < ActiveJob::Base
+  queue_as :batch_actions
+
+  BATCH_SIZE = 1
+
+  def perform(model_class, sql_query, changes)
+    total_count = model_class.constantize.count_by_sql(total_count_sql sql_query)
+    sql_query_reordered = order_by_id_sql sql_query
+
+    ((total_count.to_f / BATCH_SIZE).ceil).times do |batch_number|
```
workgena: I think if instead of `sql_query` use `scoped_collection_records` (ActiveRecord::Relation), then in this method you can use:

```ruby
scoped_collection_records.order(id: :asc).find_each(batch_size: 1) { |r| r.update!(changes) }
```
without any calculation of ordered.

---

What do you think about this proposal, could it work?

spec/jobs/async_batch_destroy_job_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'spec_helper'
+
+RSpec.describe AsyncBatchDestroyJob, type: :job do
+  describe '#perform' do
+    include_context :init_rateplan
+    include_context :init_destination, id: 1, initial_rate: 0.3
+    include_context :init_destination, id: 2, initial_rate: 0.5
+    include_context :init_destination, id: 3, initial_rate: 0.7
```
workgena: It would be nice to have more then 1000 records in this test

app/admin/billing/account.rb
Zamyatin-AA: Done

app/admin/billing/contractors.rb
Zamyatin-AA: Nice idea. Works fine.

app/jobs/async_batch_destroy_job.rb
Zamyatin-AA: Code works fine: we take first 1000 records, destroy it, then the same sql query will grab next 1000 records (because first batch already doesn't exist), destroy it, etc.
I tested it by lowering BATCH_SIZE constant to 1 or 2. 

About using Relation instead of sql_query.
We can't pass complex rails objects (like AR Relation) to delayed jobs, because those objects can't be serialized and put into job handler string. That's why we are bound to deal with pure sql (I'm not happy about that too).

spec/jobs/async_batch_destroy_job_spec.rb
Zamyatin-AA: I didn't want to overload test with too many records so I tested multiple batches case by lowering BATCH_SIZE constant.
The problem is if this case crashes we won't know it until we do the same thing again. So I agree that this test should be reworked.

It can be a good idea to mock BATCH_SIZE constant. Seems like killing two birds with one stone.

Zamyatin-AA:
@workgena 
Made some fixes and stubbed BATCH_SIZE constant in tests like that:
```
    before :each do
      stub_const('AsyncBatchDestroyJob::BATCH_SIZE', 2)
    end
```
to test multiple batches case without creating too many records.

workgena:
@Zamyatin-AA as for me(and @dmitry-sinina) all is fine. Is it possible to squash into one commit?

Zamyatin-AA:
@workgena Yes, it'd definitely better to squash commits.

PR #185, opened 2018-01-14
Use Rails-logger with YetisNode and JRPC

workgena:
JRPC-gem can show some debuggin information
for each http-jrpc-request
To see this more detailed logs in production
edit `config/environments/production.rb#L43`

```ruby
config.log_level = :debug
```


workgena:
@dmitry-sinina this PR is not very useful now. But has some debugging info in logs. I think it can be merged safely.

PR #195, opened 2018-01-22
Pass current_user to delayed jobs to store it in audit logs

stekatk:
Audit logs now keep information about ip and account of user who did batch update/delete.

### DIFF
--- a/spec/jobs/async_batch_destroy_job_spec.rb
+++ b/spec/jobs/async_batch_destroy_job_spec.rb
@@ -13,7 +13,8 @@ RSpec.describe AsyncBatchDestroyJob, type: :job do
       stub_const('AsyncBatchDestroyJob::BATCH_SIZE', 2)
     end
 
-    let(:who_is) { create :admin_user }
+    let(:admin) { create :admin_user }
+    let(:who_is) { { whodunnit: admin.id, controller_info: {} } }
 
     context 'incorrect class_name' do
       let(:model_class) { 'Fake' }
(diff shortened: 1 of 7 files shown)


spec/jobs/async_batch_destroy_job_spec.rb
```diff
@@ -7,12 +7,14 @@
...
     include_context :init_destination, id: 3, initial_rate: 0.7
 
-    subject { described_class.new(model_class, sql_query).perform}
+    subject { described_class.new(model_class, sql_query, who_is).perform }
 
     before :each do
       stub_const('AsyncBatchDestroyJob::BATCH_SIZE', 2)
     end
 
+    let(:who_is) { create :admin_user }
```
Fivell: @stekatk  looks like who_is designed to be an id of admin_user not a object
https://github.com/yeti-switch/yeti-web/pull/195/files#diff-218fbdc3dea6c12aa4b8a084680af7d9R20

spec/jobs/async_batch_update_job_spec.rb
```diff
@@ -7,12 +7,14 @@
...
 
-    subject { described_class.new(model_class, sql_query, changes).perform}
+    subject { described_class.new(model_class, sql_query, changes, who_is).perform }
 
     before :each do
       stub_const('AsyncBatchUpdateJob::BATCH_SIZE', 2)
     end
 
+    let(:who_is) { create :admin_user }
+
```
Fivell: @stekatk same about ```whoi_is```. that I mentioned above.
Also I believe tests should be added to check who_is was  passed to Audit logs

PR #203, opened 2018-02-09
Add model `System::LuaScript`

workgena:
This PR do the following:
- create table `sys.lua_scripts`
- and corresponding model `System::LuaScripts`
- ActiveAdmin page with CRUD operations `System - Lua Scripts`
- simple validations tests `spec/models/lua_script_spec.rb`

screenshot-2018-2-9 system lua scripts
---
screenshot-2018-2-10 editing columns new name 1
---
screenshot-2018-2-9 new system lua script


workgena:
I'm interesting in this:
* Do we need "Copy" action in ActiveAdmin?
* Do we need versioning on this model?
* Do we need filter by name in ActiveAdmin?
* Do we need https://highlightjs.org ?

dmitry-sinina:
> Do we need "Copy" action in ActiveAdmin?

yes
> Do we need versioning on this model?

yes
>Do we need filter by name in ActiveAdmin?

yes
>Do we need https://highlightjs.org ?

yes

db/migrate/20180209140554_create_lua_scripts.rb
```diff
@@ -0,0 +1,19 @@
+class CreateLuaScripts < ActiveRecord::Migration
+  def up
+    execute %q{
+      CREATE TABLE sys.lua_scripts (
+        id serial PRIMARY KEY,
+        name character varying NOT NULL,
```
dmitry-sinina: name should be unique

workgena:
Done all requested changes. See updated screenshots above.
@dmitry-sinina See changes in separate commit. It this issue is finished, I will squash them into single one.

dmitry-sinina:
Seems highlighting enabled only for Show page. Is it possible to enable it for Form input?

PR #226, opened 2018-02-27
Add spec for CustomersAuthNormalized.column_names

workgena:
Test if colums of CustomersAuthNormalized match CustomerAuth columns

The missing column is `external_id` and test fails.

@dmitry-sinina, you are now working with routing. Do you want add missing column by yourself or should I do it in this PR?


workgena:
did rebase with master

PR #205, opened 2018-02-12
Customers Auth multiple match conditions

workgena:
- create table `customers_auth_denormalized`
- add new fields to `customers_auth` table:
  * ips
  * src_prefixes
  * dst_prefixes
  * uri_domains
  * from_domains
  * to_domains
  * x_yeti_auths

Checklist:
- [x] write migration
- [x] update model and spec
- [x] Write test on failing situations
- [x] ActiveAdmin should use new attributes instead of old
- [x] ActiveAdmin should filtering by new attributes
- [x] ActiveAdmin show de-normalized copies (readonly)
- [x] Use new attributes in AdminAPI
- [x] Use new attributes in CustomerAPI
- [x] Update APIs documentation

This are global issues, wre broken a long time ago. Will be fixed in 3 saparate PRs:
- [ ] ~~Rework ActiveAdmin Copy-action~~
- [ ] ~~Export with new attributes~~
- [ ] ~~Import with new attributes~~

closes #204

PR #229, opened 2018-03-01
Fix "ats_as_clone"-action in ActiveAdmin resources

workgena:
Some attributes were not copied correctly(Array, Inet)

PR #129 fixed Inet attribute copy, but broke "Array of strings"
PR #205 has another approach to fix various errors with Inet type

`lib/active_record/connection_adapters/postgresql/oid/inet.rb`

```ruby
module ActiveRecord
  module ConnectionAdapters
    module PostgreSQL
      module OID # :nodoc:
        class Inet < Type::String # :nodoc:
          def type
            :inet
          end
        end
      end
    end
  end
end
```

Due to this patch, the PR #129 is redundant.
Reverting it fixes `acts_as_clone` for every attribute-type,
including "Array of Strings" and Inet

Current PR has test `copy_customers_auth_spec.rb` which covers
copy of all attribute types


workgena:
@Fivell this is controversial decision(patch inet.rb). But it seams to work well in all scenarios. @dmitry-sinina sad we use Inet as String almost everywhere.

And the problem:

```
self.ip => #<IPAddr: IPv4:0.0.0.0/0.0.0.0> .to_s => '0.0.0.0/0' #when read from database
self.ip => #<IPAddr: IPv4:0.0.0.0/0.0.0.0> .to_s => '0.0.0.0'   #when read with acts_as_clone
```

were present in some scenarios as well, not only in Copy-action.

Originally I wanted to use [attribute type cast](http://edgeapi.rubyonrails.org/classes/ActiveRecord/Attributes/ClassMethods.html), but couldn't make it work with "Array of Strings", despite of documentation. It seems `array: true` works only in Rails 5.

Fivell:
@workgena , lets try this

PR #236, opened 2018-03-04
Tag fixes

dmitry-sinina:


workgena:
@dmitry-sinina what about `db/seeds/main/class4.sql`? New `disconnect_code`-s should be there as well.

PR #237, opened 2018-03-05
remove pgq_ext

dmitry-sinina:


workgena:
Task is "Remove method perform_event". But...

`perform_batch` call `perfomr_events` then call `perform_event`, I doubt that we can remove it

/cc @Fivell

Fivell:
@workgena  sorry don't understand what are you trying to achieve

PR #238, opened 2018-03-05
Remove column `routing_tag_id` fom models

workgena:
- [x] Dialpeer
- [x] Destination
- [x] RoutingTagDetectionRule
- [x] CDR


workgena:
@dmitry-sinina please take a look at [app/models/rateplan.rb#L58](https://github.com/yeti-switch/yeti-web/blob/43d1f4f1221aa73f11e3f94e5a34aaac45fcf6b7/app/models/rateplan.rb#L58) Can you please help with this SQL?

workgena:
Rateplan SQL changed according to @dmitry-sinina solution

workgena:
Need to drop column `routing_tag_id` from CDR, but has some troubles with Postgres. Will check it tomorrow.

PR #259, opened 2018-03-16
Add `routing_tag_ids` to RoutingTagDetectionRule

workgena:
Add column `RoutingTagDetectionRule#routing_tag_ids` (Array)
it behaves the same way as in Destination/Dialpeer

Fix title of column `tag_action_value` in ActiveAdmin resources

rtdr_show


workgena:
numberlist_show

numberlist_item_show

workgena:
New column added with new migration.

PR #262, opened 2018-03-20
WIP: admin API CDRs exports endpoint

gigorok:


app/jobs/worker/cdr_export_job.rb
```diff
@@ -0,0 +1,45 @@
...
+    rescue => _e
+      #todo: when failed?
+      cdr_export.update!(status: 'Failed')
+    end
+
+    private
+
+    #todo: check on ransack availability
+    def raw_sql
+      Cdr::Cdr.select(cdr_export.fields.join(', ')).order('time_start desc').ransack(cdr_export.filters).result.to_sql
```
Fivell: it should be responsibility of CdrExport class, it should be possible to use  different cdr_export object that can have different raw_sql data

app/models/cdr_export.rb
```diff
@@ -0,0 +1,52 @@
...
+  #any cdr columns
+  ALLOWED_FIELDS = Cdr::Cdr.column_names.freeze
+
+  validates_presence_of :status, :fields, :filters
+  validate do
+    # todo add requires at time_start filter
+  end
+  validate do
+    extra_fields = fields - ALLOWED_FIELDS
+    if extra_fields.any?
```
Fivell: Imo there should be 1 abstract Export Class with minimal validation logic, to extend it and reuse with different logic

app/models/cdr_export.rb
```diff
@@ -0,0 +1,52 @@
...
+#
+# Table name: cdr_exports
+#
+#  id           :integer          not null, primary key
+#  status       :string           not null
+#  fields       :string           default([]), is an Array
+#  filters      :json             not null
+#  callback_url :string
+#  created_at   :datetime
+#  updated_at   :datetime
```
Fivell: external_id uuid can be used for callback_url integration
when export ready system notify with

POST callback_url with external_id in body or something like this

app/models/cdr_export.rb
gigorok: why we need this overhead at this level?

app/models/cdr_export.rb
gigorok: could you give me some details? we were discussing with @dmitry-sinina that GET will be enough & passing ID of export record in params to notify the client

app/models/cdr_export.rb
Fivell: @gigorok 
>> to extend it and reuse with different logic

app/models/cdr_export.rb
Fivell: @gigorok  3d party system has it's own ids , uuids etc,

gigorok: yes, thats why after making POST request from 3d party system to yeti admin API (cdr_exports endpoint) we received export ID which have to be stored on 3d party database. in this case we can match it when callback_url will be touched

Fivell:
Such behavior is better , agree

20 марта 2018 г., 23:27 +0200, Igor Gonchar <notifications@github.com>, писал:

app/jobs/worker/cdr_export_job.rb
```diff
@@ -0,0 +1,45 @@
...
+    def dir_path
+      Rails.configuration.yeti_web.fetch('cdr_export').fetch('dir_path').chomp('/')
+    end
+
+    def filename
+      cdr_export.id
+    end
+
+    def find_cdr_export(id)
+      @cdr_export = CdrExport.find(id)
```
senid231: at line 8 it already assigned to `@cdr_export`

app/jobs/worker/cdr_export_job.rb
```diff
@@ -0,0 +1,45 @@
+module Worker
+  class CdrExportJob < ActiveJob::Base
+    queue_as 'cdr_export'
+
+    attr_reader :cdr_export
+
+    def perform(cdr_export_id)
+      @cdr_export = find_cdr_export(cdr_export_id)
+
+      Cdr::Cdr.connection.execute("COPY (#{cdr_export.export_sql}) TO '#{file_path}' WITH CSV HEADER;")
```
senid231: you need instance variable `@cdr_export` for file_path
why not just add method
```ruby
def file_path_for(cdr_export)
  "#{dir_path}/#{cdr_export.id}.csv"
end
```
an use it without unnecessary instance variables like this:
```ruby
cdr_export = find_cdr_export(cdr_export_id)
file_path = file_path_for(cdr_export)
Cdr::Cdr.connection.execute("COPY (#{cdr_export.export_sql}) TO '#{file_path}' WITH CSV HEADER;")
```

app/models/cdr_export.rb
```diff
@@ -0,0 +1,70 @@
...
+    'time_start_gteq',
+    'success_eq',
+    'customer_auth_external_id_eq',
+    'failed_resource_type_id_eq',
+    'src_prefix_in_contains',
+    'dst_prefix_in_contains',
+    'src_prefix_routing_contains',
+    'dst_prefix_routing_contains',
+    'customer_acc_external_id_eq',
+    'is_last_cdr_eq'
```
senid231: maybe elements of array should be frozen as well

config/routes.rb
```diff
@@ -80,6 +80,7 @@
           namespace :cdr do
             jsonapi_resources :cdrs, only: [:index, :show] do
             end
+            jsonapi_resources :cdr_exports, only: [:create]
```
senid231: do we have an ability to retrieve completed export file through the API?

db/migrate/20180320120746_create_cdr_exports.rb
```diff
@@ -0,0 +1,12 @@
+class CreateCdrExports < ActiveRecord::Migration
+  def change
+    create_table 'sys.cdr_exports' do |t|
+      t.string :status, null: false
+      t.string :fields, default: [], array: true
```
senid231: maybe `null: false` needed too?

config/routes.rb
gigorok: later, in this step only create allowed

PR #275, opened 2018-03-27
admin ui exports

gigorok:
you can track of CDR exports, download already dumped CSV files, create new CDR exports through admin panel

### DIFF
--- a/app/models/cdr_export.rb
+++ b/app/models/cdr_export.rb
@@ -19,10 +19,19 @@ class CdrExport < Yeti::ActiveRecord
   STATUS_PENDING = 'Pending'.freeze
   STATUS_COMPLETED = 'Completed'.freeze
   STATUS_FAILED = 'Failed'.freeze
+  STATUSES = [
+    STATUS_PENDING,
+    STATUS_COMPLETED,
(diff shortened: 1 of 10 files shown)


app/models/cdr_export.rb
```diff
@@ -19,10 +19,20 @@ class CdrExport < Yeti::ActiveRecord
...
+
+  #need for activeadmin form
+  attr_accessor :customer_acc_id_eq,
+    :is_last_cdr_eq, :time_start_gteq, :time_start_lteq
 
   validates_presence_of :status, :fields, :filters
   validate do
-    if filters.keys.exclude?('time_start_lteq') || filters.keys.exclude?('time_start_gteq')
+    if filters.keys.exclude?('time_start_lteq') || filters.keys.exclude?('time_start_gteq') ||
+        filters['time_start_gteq'].empty? || filters['time_start_lteq'].empty?
```
Fivell: @gigorok  I think it can be simplier just
 ```filters['time_start_gteq'].empty? || filters['time_start_lteq'].empty?```

gigorok: agree

PR #291, opened 2018-04-08
upgrade to rails 5.0

gigorok:


dmitry-sinina:
After run
```
 bundle exec rake db:migrate
 bundle exec rake db:second_base:migrate
```
I found that db/secondbase/structure.sql and db/structure.sql was changed:
```
+CREATE TABLE gui.ar_internal_metadata (
+    key character varying NOT NULL,
+    value character varying,
+    created_at timestamp without time zone NOT NULL,
+    updated_at timestamp without time zone NOT NULL
+);

+--
+-- Name: ar_internal_metadata; Type: TABLE; Schema: cdr; Owner: -; Tablespace: 
+--
+
+CREATE TABLE cdr.ar_internal_metadata (
+    key character varying NOT NULL,
+    value character varying,
+    created_at timestamp without time zone NOT NULL,
+    updated_at timestamp without time zone NOT NULL
+);
```
Seems such changes of structure should be commited. Also I think we don't need such table at CDR database.

config/environments/production.rb
```diff
@@ -1,78 +1,85 @@
...
+  # config.action_cable.allowed_request_origins = [ 'http://example.com', /http:\/\/example.*/ ]
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   # config.force_ssl = true
 
-  # See everything in the log (default is :info)
-  config.log_level = :info
+  # Use the lowest log level to ensure availability of diagnostic information
+  # when problems arise.
+  config.log_level = :debug
```
Fivell: @gigorok what is the reason of this?

config/initializers/cookies_serializer.rb
```diff
@@ -0,0 +1,5 @@
+# Be sure to restart your server when you modify this file.
+
+# Specify a serializer for the signed and encrypted cookie jars.
+# Valid options are :json, :marshal, and :hybrid.
+Rails.application.config.action_dispatch.cookies_serializer = :marshal
```
Fivell: @gigorok use : hybrid here

spec/controllers/api/rest/admin/cdr/cdr_exports_controller_spec.rb
```diff
@@ -86,7 +86,7 @@
         expect(response.status).to eq(422)
         expect(JSON.parse(response.body)['errors']).to match_array(
           hash_including(
-            'detail' => 'filters - unknown_filter not allowed'
+            'detail' => 'filters - unknown-filter not allowed'
```
Fivell: @gigorok  why this introduced on rails 4-5 migration commit? do not understand how is related

config/environments/production.rb
gigorok: this changes will be suggested after running the `rails app:update` command. do you think we have to leave `info` ?

config/initializers/cookies_serializer.rb
gigorok: agree, we can use `hybrid` to move to `json` in the future

spec/controllers/api/rest/admin/cdr/cdr_exports_controller_spec.rb
gigorok: this change is related with upgrading of `jsonapi-resources` from 0.9.1.beta1 to 0.9.1.beta2. looks like this PR https://github.com/cerebris/jsonapi-resources/pull/1069

config/environments/production.rb
Fivell: @gigorok personally don't think we should change it without reason,  I even think info is too much or production and warn should be used, @dmitry-sinina  what do you think?

spec/controllers/api/rest/admin/cdr/cdr_exports_controller_spec.rb
Fivell: @gigorok  thanks, good to know,  jsonapi-resources update via separate commit would be perfect

gigorok: @Fivell I was wrong. nothing should be changed in this case. 0.9.1.beta2 deal with filters on index action, we are performing custom validation error. anyway I rolled back jsonapi-resources to 0.9.1.beta1

PR #292, opened 2018-04-10
Patch ActiveAdminImport method batch_slice_columns

workgena:
Add patch of `active_admin_import` gem with method
`batch_slice_columns`. This method allows CSV to has redundant
or database-table mismatch columns.

We can not update `active_admin_impor` because it depends
on `activeadmin 1.0.0`. And `yeti-web` still uses patched `activeadmin`

PS: when PR#291 will be merged, we can drop this patch and simply update `active_admin_import` gem.


lib/active_admin_import/importer.rb
```diff
@@ -0,0 +1,21 @@
+module ActiveAdminImport
+  class Importer
+
+    # REMOVE this after update of `active_admin_import` and `activeadmin` v1
+    # and remove correcponding line in `config/initializers/yeti.rb`
+    def batch_slice_columns(slice_columns)
+      use_indexes = []
+      headers.values.each_with_index do |val, index|
+        use_indexes << index if val.in?(slice_columns)
+      end
```
Fivell: @workgena  feel free to add this to master of active_admin_import why patching ?

workgena:
Added(commit --amend) new behavior: all foreign_key columns are being ignored.

This code do the trick:

```ruby
skip_columns = options[:resource_class].reflect_on_all_associations(:belongs_to).map(&:foreign_key)
```

workgena:
(commit --amend) Added `skip_columns` to Destination, Dealpeer with `routing_tag_ids` and CustoemrsAuth with `tag_action_value`.

PR #304, opened 2018-04-27
Change display of column routing_tag_mode

workgena:
For Dialpeer, Destination and RoutingTagDetectionRule
change the way of display `routing_tag_mode`-column:

1. Remove column `routing_tag_mode` from index page
2. Change display of `routing_tags` column:
  * **UA & US** --> when mode is AND
  * **UA | US** --> when mode is OR

default


workgena:
Ready to merge.

Issue #306, opened 2018-05-12
remote_stats/cdrs_summary.json is broken


dmitry-sinina:
```
[...]

PR #312, opened 2018-05-14
refs #306, apply default cdrs sorting

gigorok:
default_scope is evil. try to prevent yourself from using it


gigorok:
@workgena, thanks. also should be added sorting explicitly

workgena:
+ API/Customer/CDR

PR #319, opened 2018-05-31
Rewrite `cdr.cdr` table partitioning to Ruby

workgena:
Rewrite CDR table partitioning from SOL-functions to Ruby-code.

- [x] Full test coverage for CDR partitioning
- [x] Implement CDR-partition creation with Ruby code
- [x] Remove old PG-function with migration

### DIFF
--- a/app/models/cdr/table.rb
+++ b/app/models/cdr/table.rb
@@ -12,12 +12,14 @@
 #
 
 class Cdr::Table < Cdr::Base
+  extend PgCdrPartitioning
 
   self.table_name = 'sys.cdr_tables'
 
-
   has_paper_trail class_name: 'AuditLogItem'
-  
+
+  scope :active, -> { where(active: true) }
+
   def display_name
     "#{self.name}"
   end
@@ -32,13 +34,13 @@ class Cdr::Table < Cdr::Base
       _reload_trigger
     end
   end
-  
+
   def unload
     self.execute_sp("SELECT sys.cdr_export_data(?,?)" , self.name, GuiConfig.cdr_unload_dir)
   end
 
   def _reload_trigger
-    self.execute_sp("SELECT sys.cdrtable_tgr_reload()")
+    self.reload_cdr_i_tgf
   end
 
   def archive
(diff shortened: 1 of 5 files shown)


app/models/cdr/table.rb
```diff
@@ -53,15 +53,59 @@ def archive
...
-      self.execute_sp("SELECT * FROM sys.cdr_createtable(1)")  ##try to create table for current month
-      self.execute_sp("SELECT * FROM sys.cdr_createtable(0)") ##for next month
+      today = Date.today
+      create_table_for(today.prev_month)
+      create_table_for(today)
+      create_table_for(today.next_month)
+    end
+  end
+
+  def self.create_table_for(date_from)
```
workgena: @dmitry-sinina please check this out.

Do we need to rewrite this functions, as well?
- sys.cdr_reindex
- sys.cdrtable_tgr_reload

app/models/cdr/table.rb
```diff
@@ -53,15 +55,40 @@ def archive
...
-      self.execute_sp("SELECT * FROM sys.cdr_createtable(0)") ##for next month
+      today = Date.today
+      # TODO: find out what design pattern suits here
+      Yeti::PgTablePartition.new(connection: connection, date_from: today.prev_month).create_partition
+      Yeti::PgTablePartition.new(connection: connection, date_from: today).create_partition
+      Yeti::PgTablePartition.new(connection: connection, date_from: today.next_month).create_partition
     end
   end
 
+  def self.destroy_partitions
```
dmitry-sinina: This method should raise exception if application running in production mode. It is not save to allow drop all data in production

dmitry-sinina:
> Should we remove related PG-function in this PR?

yes

workgena:
I've done some refactoring. The `PgCdrPartitioning` will be easy to convert into generic `PgPartitioning`, later, if this implementations will work fine :floppy_disk: :balloon:

@dmitry-sinina please watch through the changes. If it is OK, then I'll squash commits into one, and then merge.

workgena:
squashed on to one commit, ready to merge

PR #324, opened 2018-06-21
PgTablePartitioning module

workgena:
Implement `PgPartitioningMixin` which allows to make any table partition-able.

Usage described in module itself.

This PG includes migration of making partitions for `auth_log.auth_log` table. But without actual creation of them(no job scheduled). It means then auth_log will not work at all.

Before merging Pull Request, @dmitry-sinina should to decide:
- add job for `auth_log` partitioning
or
- delete migration(make only the example of it)


app/models/cdr/table.rb
```diff
@@ -12,11 +12,18 @@
...
 class Cdr::Table < Cdr::Base
-  extend PgCdrPartitioning
-
   self.table_name = 'sys.cdr_tables'
 
-  has_paper_trail class_name: 'AuditLogItem'
+  include PgPartitioningMixin
+
+  self.partitioned_model = Cdr::Cdr
+  self.partition_schema = 'cdr'.freeze
```
workgena: Give default value to if, if not defined. Take the same schema as `partitioned_model` is.

app/models/cdr/table.rb
```diff
@@ -12,11 +12,18 @@
...
+  self.partition_key = :time_start
+  self.trigger_function_name = 'cdr.cdr_i_tgf'.freeze
+  self.trigger_name = 'cdr.cdr_i_tg'.freeze
```
workgena: Let it be with default values, generate them from `self.partitioned_mode.table_name`

Issue #362, opened 2018-09-05
add possibility to remove cdrs exports


gigorok:
add delete_url to yeti_web.yml (section cdr_export)

while deleting of cdrs export through API or admin panel create dj which make DELETE request to delete_url/<cdrs_export ID>. if returned 200 or 404 then change status to removed. if http code is not 200 or 404 dj should exists with error.

deleting of files will be performed by nginx & issue will be delegated to @dmitry-sinina later

PR #363, opened 2018-09-05
refs #362, removing cdrs exports

gigorok:
see https://github.com/yeti-switch/yeti-web/issues/362


app/admin/cdr/cdr_exports.rb
```diff
@@ -34,6 +36,12 @@
     render body: nil
   end
 
+  member_action :delete, method: :delete do
+    Worker::RemoveCdrExportFileJob.perform_later(resource.id)
+    flash[:notice] = 'The file will be deleted in async way!'
```
Fivell: @gigorok in this situation multiple delete possible, deletion should 
1) change status
2) create dj to delete file

PR #389, opened 2018-10-21
Test invoice index and new page

workgena:
This is a example of "How to write tests for New and Index pages".

Take an look at `spec/features/invoices/new_invoice_feature_spec.rb`

0. This feature-test requires JavaScript to be enabled `js: true`. Because this form has "chosen"-select with AJAX loading of Accounts. Try avoid `js: true` because it is slow... :snail: ...
1. In `before` block we prepare database records needed for New-form.
2. `include_context :fill_form, 'new_billing_invoice'`

    The "new_billing_invoice" is the HTML-id of a form: `<form id="new_billing_invoice">`
3. `let(:attributes) do` is a key-value structure, where key is a HTML id of and input-tag (excluding form-prefix).
    For example: `<input id="billing_invoice_start_date" name="billing_invoice[start_date]" ...>`
    the ID would be `start_date` taken from "billing_invoice_start_date" string.
4. The test case `it 'creates new invoice succesfully'` performs form submission and check if page has the success-massage: `Invoice was successfully created`.
    Then, if you with, you can take newly created record(Billing::Invoice.last) and validate each attribute. Or just check if it was created, like this: `expect(Billing::Invoice.count).to eq(1)`

### DIFF
--- a/lib/active_admin/views/footer.rb
+++ b/lib/active_admin/views/footer.rb
@@ -2,12 +2,10 @@ module ActiveAdmin
   module Views
     class Footer < Component
 
-     private
+      private
 
       def powered_by_message
-
-        para "Copyright #{Date.today.year} Yeti Admin ver: #{Yeti::Application.config.app_build_info.fetch('version', 'unknown')}. Routing ver #{Yeti::ActiveRecord::DB_VER}. CDR ver #{Cdr::Base::DB_VER}.".html_safe
-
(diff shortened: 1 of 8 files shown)


workgena:
also take a look at [create_destination_spec.rb](https://github.com/yeti-switch/yeti-web/blob/66c5f62ecbd4c696b65dc953c49c64ef0758796b/spec/features/destinations/create_destination_spec.rb). The similar test but without JavaScript.

PR #418, opened 2018-12-17
#415 pgq processor exit code 1 on error

### LINKED ISSUE #415: Pgq processors exit code
```
  def self.check_interrupted(file, line)
    if self.interrupted?
        logger.info  "[#{file}: #{line}] Worker has been halted with #{self.interrupted}"
        exit(0)
    end
  end
```
We should return non-zero exit code on error.

senid231:
fix #415

### DIFF
--- a/pgq-processors/ext/pgq/lib/pgq/worker.rb
+++ b/pgq-processors/ext/pgq/lib/pgq/worker.rb
@@ -2,10 +2,17 @@ class Pgq::Worker
   attr_reader :logger, :class, :sleep_time
 
   cattr_accessor :interrupted do false end
+  cattr_accessor :graceful_exit do true end
   cattr_accessor :logger
 
   def self.shutdown!(signal)
-     self.interrupted = signal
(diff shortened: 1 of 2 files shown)


pgq-processors/ext/pgq/lib/pgq/worker.rb
```diff
@@ -2,10 +2,12 @@ class Pgq::Worker
   attr_reader :logger, :class, :sleep_time
 
   cattr_accessor :interrupted do false end
+  cattr_accessor :interrupted_with_exception do false end
   cattr_accessor :logger
 
-  def self.shutdown!(signal)
-     self.interrupted = signal
+  def self.shutdown!(signal, with_exception = false)
+    self.interrupted = signal
```
Fivell: @senid231 maybe simplify it and leave 1 argument but check if exception inside shutdown! and set interrupted_with_exception as private

senid231: done

PR #430, opened 2019-01-10
upgrade rails to 5.2

gigorok:
1. upgrade activeadmin to latest version
2. use jquery3 instead of jquery2
3. add bootsnap gem to speedup booting
4. remove unused css
5. etc (look at changes)


app/assets/stylesheets/active_admin.css.scss
```diff
@@ -22,6 +22,7 @@
 @import "font-awesome";
 @import "active_admin/components/active_admin_scoped_collection_actions";
 @import "vendor/highlightjs/default";
+@import "yeti/yeti.css"
```
workgena: This might not work in production-mode. Better check in production mode before merge. 

Recently I've had a problem - CSS files would not be imported. In case of such error, the solution is to rename files `*.css` to `*.scss`.

gigorok: thanks. I'll check it

workgena:
I've did a small check. Unfortunately we do not covered all pages with system tests. So some pages may be broken, even if tests are green :alien:

Broken pages, i've noticed(the errors are unfamiliar to me):
* /gateways/new
* /dialpeers/new
* /audit_log_items

Also I did `assets:precompile + rails s` in production mode, all works fine(js, css).

gigorok:
@workgena what kind of errors did you receive?

workgena:
```
Started GET "/gateways/new" for 127.0.0.1 at 2019-01-15 14:55:29 +0200
Processing by GatewaysController#new as HTML
  AdminUser Load (0.3ms)  SELECT  "admin_users".* FROM "admin_users" WHERE "admin_users"."id" = $1 ORDER BY "admin_users"."id" ASC LIMIT $2  [["id", 3], ["LIMIT", 1]]
  ↳ app/controllers/application_controller.rb:45
Completed 500 Internal Server Error in 2ms

ArgumentError (wrong number of arguments (given 0, expected 1)):
  
config/initializers/instrumentation_notification.rb:28:in `block in process_action'
config/initializers/instrumentation_notification.rb:27:in `process_action'
```

gigorok:
> ```
> Started GET "/gateways/new" for 127.0.0.1 at 2019-01-15 14:55:29 +0200
> Processing by GatewaysController#new as HTML
>   AdminUser Load (0.3ms)  SELECT  "admin_users".* FROM "admin_users" WHERE "admin_users"."id" = $1 ORDER BY "admin_users"."id" ASC LIMIT $2  [["id", 3], ["LIMIT", 1]]
>   ↳ app/controllers/application_controller.rb:45
> Completed 500 Internal Server Error in 2ms
> 
> ArgumentError (wrong number of arguments (given 0, expected 1)):
>   
> config/initializers/instrumentation_notification.rb:28:in `block in process_action'
> config/initializers/instrumentation_notification.rb:27:in `process_action'
> ```

related with upgrading of activeadmin. fixed

PR #435, opened 2019-01-17
Fix policy detection when omitting action or subject

senid231:


senid231:
@dmitry-sinina fixed

senid231:
@dmitry-sinina rubocop fixed

PR #448, opened 2019-02-26
added comments policy

segrey11:
It is impossible to add comment to contractor due to missing policy class for Comments controller


segrey11:
Thanks for the advice. Test added.

PR #453, opened 2019-03-15
Feature. Ransack filters for admin API

max-konin:
**TODO**

- [ ] Add filters by field with array type
- [x] Add filters by field with uuid type
- [x] Add filters by field with datetime type
- [ ] Add filters by field with json type (?)
- [x] Add filters by field with inet type

### DIFF
--- a/app/resources/base_resource.rb
+++ b/app/resources/base_resource.rb
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 
 class BaseResource < JSONAPI::Resource
-  AVAILABLE_FILTERS_TYPES = %i[string number boolean].freeze
   RANSACK_TYPE_SUFIXES_DIC = {
-    string: %w[eq not_eq matches],
     boolean: %w[eq not_eq],
-    number: %w[eq not_eq gt gteq lt lteq]
+    datetime: %w[eq not_eq gt gteq lt lteq],
+    number: %w[eq not_eq gt gteq lt lteq],
+    string: %w[eq not_eq matches]
   }.freeze
 
   abstract
@@ -16,7 +16,7 @@ class BaseResource < JSONAPI::Resource
 
   def self.ransack_filter(attr, options)
     type = options[:type]
-    raise ArgumentError, "type #{type} is not supported" unless AVAILABLE_FILTERS_TYPES.include?(type)
+    raise ArgumentError, "type #{type} is not supported" unless RANSACK_TYPE_SUFIXES_DIC.key?(type)
 
     RANSACK_TYPE_SUFIXES_DIC[type].each do |suf|
       ransack_operator = "#{attr}_#{suf}"
--- a/app/resources/base_resource.rb
+++ b/app/resources/base_resource.rb
@@ -5,7 +5,7 @@ class BaseResource < JSONAPI::Resource
     boolean: %w[eq not_eq],
     datetime: %w[eq not_eq gt gteq lt lteq],
     number: %w[eq not_eq gt gteq lt lteq],
-    string: %w[eq not_eq matches],
+    string: %w[eq not_eq cont start end],
     uuid: %w[eq not_eq]
   }.freeze
 
--- a/app/resources/base_resource.rb
+++ b/app/resources/base_resource.rb
@@ -1,9 +1,29 @@
 # frozen_string_literal: true
 
 class BaseResource < JSONAPI::Resource
+  AVAILABLE_FILTERS_TYPES = %i[string number boolean].freeze
+  RANSACK_TYPE_SUFIXES_DIC = {
+    string: %w[eq not_eq matches],
+    boolean: %w[eq not_eq],
+    number: %w[eq not_eq gt gteq lt lteq]
+  }.freeze
+
   abstract
 
   def self.type(custom_type)
     self._type = custom_type
+  end
+
+  def self.ransack_filter(attr, options)
+    type = options[:type]
+    raise ArgumentError, "type #{type} is not supported" unless AVAILABLE_FILTERS_TYPES.include?(type)
+
+    RANSACK_TYPE_SUFIXES_DIC[type].each do |suf|
+      ransack_operator = "#{attr}_#{suf}"
+
+      filter ransack_operator, apply: (lambda do |records, value, _options|
+        records.ransack(ransack_operator => value.first).result
+      end)
+    end
   end
 end
--- a/app/resources/base_resource.rb
+++ b/app/resources/base_resource.rb
@@ -1,15 +1,6 @@
 # frozen_string_literal: true
 
 class BaseResource < JSONAPI::Resource
-  RANSACK_TYPE_SUFIXES_DIC = {
-    boolean: %w[eq not_eq],
-    datetime: %w[eq not_eq gt gteq lt lteq in not_in],
-    inet: %w[eq not_eq in not_in],
-    number: %w[eq not_eq gt gteq lt lteq in not_in],
-    string: %w[eq not_eq cont start end in not_in cont_any],
-    uuid: %w[eq not_eq in not_in]
-  }.freeze
-
   abstract
 
   def self.type(custom_type)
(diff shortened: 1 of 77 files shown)


app/resources/base_resource.rb
```diff
@@ -1,9 +1,29 @@
 # frozen_string_literal: true
 
 class BaseResource < JSONAPI::Resource
+  RANSACK_TYPE_SUFIXES_DIC = {
+    boolean: %w[eq not_eq],
+    datetime: %w[eq not_eq gt gteq lt lteq],
+    number: %w[eq not_eq gt gteq lt lteq],
+    string: %w[eq not_eq matches]
```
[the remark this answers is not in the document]
max-konin: It makes sense, thx

app/resources/base_resource.rb
```diff
@@ -1,9 +1,31 @@
...
+    type = options[:type]
+    raise ArgumentError, "type #{type} is not supported" unless RANSACK_TYPE_SUFIXES_DIC.key?(type)
+
+    RANSACK_TYPE_SUFIXES_DIC[type].each do |suf|
+      ransack_operator = "#{attr}_#{suf}"
+
+      filter ransack_operator, apply: (lambda do |records, value, _options|
+        records.ransack(ransack_operator => value.first).result
+      end)
+    end
```
senid231: there is one thing about `eq` filter
by providing ransack `filtername_eq` it instead of default `JSONAPI` `filtername` filter we remove one useful feature - filtering by one of multiple values provided
you no longer can't use filter like `?filter[src_prefix]=123,456` to receive records which matches any of these `src_prefix` values

I suggest to extend `ransack_filter` functional with something like this:
```ruby
  RANSACK_TYPE_SUFIXES_DIC = {
    boolean: %w[eq not_eq],
    datetime: %w[eq not_eq gt gteq lt lteq],
    inet: %w[eq not_eq],
    number: %w[eq not_eq gt gteq lt lteq],
    string: %w[eq not_eq cont start end],
    uuid: %w[eq not_eq]
  }.freeze

  RANSACK_TYPE_MULTIPLE_SUFIXES_DIC = {
    string: { eq: :in, not_eq: :not_in, cont: :cont_any }.freeze,
    datetime: { eq: :in, not_eq: :not_in }.freeze,
    inet: { eq: :in, not_eq: :not_in }.freeze,
    number: { eq: :in, not_eq: :not_in }.freeze,
    uuid: { eq: :in, not_eq: :not_in }.freeze
  }.freeze

  def self.ransack_filter(attr, options)
    type = options[:type]
    raise ArgumentError, "type #{type} is not supported" unless RANSACK_TYPE_SUFIXES_DIC.key?(type)

    RANSACK_TYPE_SUFIXES_DIC[type].each do |suf|
    ransack_operator = "#{attr}_#{suf}"
    ransack_multiple_operator =  RANSACK_TYPE_MULTIPLE_SUFIXES_DIC.fetch(type, {})[suf.to_sym]

    filter ransack_operator, apply: -> do |records, values, _options|
      if values.size > 1 && ransack_multiple_operator
        records.ransack(ransack_multiple_operator => values).result
      else
        records.ransack(ransack_operator => values.join(',')).result
      end
    end
  end
```

max-konin: @senid231  may be easier to add in, not_in and cont_any operators to RANSACK_TYPE_SUFIXES_DIC?

senid231: @max-konin we can do that too.
but we should rewrite logic of ransack filter applier, because line below will not work
`records.ransack(ransack_operator => value.first).result`

currently we 2 types of filters
filters that accepts array with one or more elements 
and filters that does accepts only one value

I think we need to verify filter value correctly

for example:

* for type `boolean` value can be only `true` or `false` - all other values should respond with `400 bad request: invalid filter value`
* for type `string` filters `eq not_eq cont start end` value can't be array too but value can contain comma (in this case `jsonapi-resources` will split value into array) - so we need to restore original value
* for type `uuid` filters `eq not_eq` value can't be array and should respond with `400`
* for type `inet` filters `eq not_eq` value can't be array and should respond with `400`
* for type `number` filters `eq not_eq gt gteq lt lteq` value can't be array and should respond with `400`
* for type `datetime` filters `eq not_eq gt gteq lt lteq` value can't be array and should respond with `400`

for filters  `in` and `not_in`, `cont_any` value can be array with one or more elements
filters `in` and `not_in` can't be added for `boolean` type
filter `cont_any` can be added only for `string` type

senid231: all filters those filters should respond with `400` if value is empty

senid231: it would be good to add value verifying according to filter type - but it's task for different PR

max-konin: > but we should rewrite logic of ransack filter applier, because line below will not work
records.ransack(ransack_operator => value.first).result

why? If we pass an array as a value. the value is 2dim array. For example: `?filter[name][]="one"&filter[name][]="two"`  value in filter bloc - `[["one", "two"]]`

senid231: @max-konin by default `jsonapi-resources` split filter value by comma

so if you want to filter by two values you should send query `filter[name]=one,two` and value in the `proc` will be `['one', 'two']`

senid231: value `[["one", "two"]]` looks like a bug for me

senid231: passing array to ransack filter that doesn't support it looks like a bug too
```ruby
User.ransack(name_eq: ['one', 'two']).result.to_sql
 => "SELECT \"users\".* FROM \"users\" WHERE \"users\".\"name\" = '[\"one\", \"two\"]'"
```
you will not receive result that you've waiting for

max-konin: @senid231  yes. I see. I tried to pass filter as an array. a string with commas is ok. nevertheless, what decision will we take? Separate filters for array (attr_name_in, attr_name_not_in) OR your solution with RANSACK_TYPE_MULTIPLE_SUFIXES_DIC?.

senid231: I think filters for array will be simpler. But with additional verification
something like:
```ruby
  RANSACK_TYPE_SUFFIXES_DIC = {
      boolean: %w[eq not_eq],
      datetime: %w[eq not_eq gt gteq lt lteq in not_in],
      inet: %w[eq not_eq in not_in],
      number: %w[eq not_eq gt gteq lt lteq in not_in],
      string: %w[eq not_eq cont start end in not_in cont_any],
      uuid: %w[eq not_eq  in not_in]
  }.freeze

  RANSACK_ARRAY_SUFFIXES = %w[in not_in cont_any]

  def self.ransack_filter(attr, options)
    type = options[:type]
    raise ArgumentError, "type #{type} is not supported" unless RANSACK_TYPE_SUFFIXES_DIC.key?(type)

    RANSACK_TYPE_SUFFIXES_DIC[type].each do |suf|
      filter_name = "#{attr}_#{suf}"
      is_array_suffix = RANSACK_ARRAY_SUFFIXES.include?(suf)
      applier = ->(records, values, _opts) do
        records.ransack(filter_name => values).result
      end
      single_verifier = ->(values, _ctx) do
        raise JSONAPI::Exceptions::InvalidFilterValue.new(filter_name, values.join(',')) if values.size != 1
        values.first
      end
      filter filter_name, apply: applier, verify: is_array_suffix ? nil : single_verifier
    end
  end
```

app/resources/base_resource.rb
```diff
@@ -1,9 +1,30 @@
...
+  def self.ransack_filter(attr, options)
+    type = options[:type]
+    raise ArgumentError, "type #{type} is not supported" unless RANSACK_TYPE_SUFIXES_DIC.key?(type)
+
+    RANSACK_TYPE_SUFIXES_DIC[type].each do |suf|
+      builder = RansackFilterBuilder.new(attr: attr, operator: suf)
+      filter builder.filter_name,
+             verify: ->(values, _ctx) { builder.verify(values) },
+             apply: ->(records, values, _opts) { builder.apply(records, values) }
+    end
```
senid231: I think `RANSACK_TYPE_SUFIXES_DIC` should be put to `RansackFilterBuilder`

then code in base controller would be

```ruby
  def self.ransack_filter(attr, options)
    type = options[:type]
    raise ArgumentError, "type #{type} is not supported" unless RansackFilterBuilder.type_supported?(type)

     RansackFilterBuilder.suffixes_for_type(type).each do |suf|
      builder = RansackFilterBuilder.new(attr: attr, operator: suf)
      filter builder.filter_name,
             verify: ->(values, _ctx) { builder.verify(values) },
             apply: ->(records, values, _opts) { builder.apply(records, values) }
    end
  end
```

and `RansackFilterBuilder` will have code
```ruby
class RansackFilterBuilder
RANSACK_TYPE_SUFIXES_DIC = {
    boolean: %w[eq not_eq],
    datetime: %w[eq not_eq gt gteq lt lteq in not_in],
    inet: %w[eq not_eq in not_in],
    number: %w[eq not_eq gt gteq lt lteq in not_in],
    string: %w[eq not_eq cont start end in not_in cont_any],
    uuid: %w[eq not_eq in not_in]
  }.freeze

  class << self
    def suffixes_for_type(type)
      RANSACK_TYPE_SUFIXES_DIC[type]
    end

    def type_supported?(type)
      RANSACK_TYPE_SUFIXES_DIC.key?(type)
    end
  end
end
```

PR #491, opened 2019-06-08
Active calls API

senid231:


senid231:
failing build on master https://travis-ci.org/yeti-switch/yeti-web/builds/543316547

I've created #494 to fix this

PR #492, opened 2019-06-08
Active calls API patch 1.8

senid231:
patch 1.8 with #491


senid231:
#494 also cherry-picked here to fix CI

PR #673, opened 2020-04-15
moved configs for bundler tool

Ivanov-Anton:
**moved bundler's configs** named "with", "without" and "deployment" from inline call as key for command bundle to config globally, 
**cause of: outputing warning** 
`" [DEPRECATED] The `--deployment` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions."`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,14 +105,19 @@ bundler:
 .PHONY: gems
 gems: bundler
 	$(info:msg=Install/Update gems)
-	$(bundle_bin) install --jobs=4 --deployment --without development test
+	$(bundle_bin) config --delete with
+	$(bundle_bin) config set without 'development test'
+	$(bundle_bin) config set deployment 'true'
+	$(bundle_bin) install --jobs=4
 	$(bundle_bin) clean
 
 
 .PHONY: gems-test
 gems-test: bundler
 	$(info:msg=Install/Update gems for tests)
-	$(bundle_bin) install --jobs=4 --deployment --with development test
+	$(bundle_bin) config set with 'development test'
+	$(bundle_bin) config set deployment 'true'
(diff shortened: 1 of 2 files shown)


dmitry-sinina:
looks like yeti-web/pgq-processors/Makefile also should be modified

Makefile
```diff
@@ -105,14 +105,19 @@ bundler:
 .PHONY: gems
 gems: bundler
 	$(info:msg=Install/Update gems)
-	$(bundle_bin) install --jobs=4 --deployment --without development test
+	$(bundle_bin) config --delete with
+	$(bundle_bin) config set without 'development test'
+	$(bundle_bin) config set development 'true'
```
gigorok, line 110: $(bundle_bin) config set deployment 'true'

Makefile
```diff
@@ -105,14 +105,19 @@ bundler:
 .PHONY: gems
 gems: bundler
 	$(info:msg=Install/Update gems)
-	$(bundle_bin) install --jobs=4 --deployment --without development test
+	$(bundle_bin) config --delete with
```
gigorok, line 108: do we need it?
 $(bundle_bin) config --delete with

Ivanov-Anton: Yes, because we get the following error: 
`a group cannot be in both `with` & `without` simultaneously`

Makefile
Ivanov-Anton: > $(bundle_bin) config set deployment 'true'

Yes you are right

Ivanov-Anton:
> @Ivanov-Anton looks like yeti-web/pgq-processors/Makefile also should be modified

Any other comments about the changes?

PR #693, opened 2020-05-06
Create missing partitions

Ivanov-Anton:
`@connection.exec_params(sql, type_casted_binds) ActiveRecord::StatementInvalid: PG::CheckViolation: ERROR:  no partition of relation "api_requests" found for row DETAIL:  Partition key of the failing row contains (created_at) = (2020-05-06 08:32:27.462837+00). : INSERT INTO "logs"."api_requests" ("created_at", "path", "method", "status", "controller", "action", "page_duration", "db_duration", "params") VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9) RETURNING "id"`

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -120,6 +120,14 @@ RSpec.configure do |config|
     DatabaseCleaner.clean_with :truncation
   end
 
+  config.before(:suite) do
+    # Create partition for current+next monthes if not exists
+    Cdr::Cdr.add_partitions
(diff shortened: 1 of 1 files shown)


dmitry-sinina:
why in spec/acceptance/rest/customer/api/v1/networks_spec.rb ? There are many tests requires this partition

dmitry-sinina:
I think it is reasonable to add partition once, when rspec stats

PR #696, opened 2020-05-08
depr warns: wraps unsafe pluck to Arel.sql

Ivanov-Anton:
cause of: outputting the following warning
```
DEPRECATION WARNING: Dangerous query method
 (method whose arguments are used as raw SQL) called with non-attribute argument(s):
```

spec for reproduce warning: 
spec/requests/api/rest/clickhouse_dictionaries/network_prefixes_spec.rb:3
spec/features/billing/accounts/index_accounts_spec.rb:8
spec/features/billing/invoices/index_invoices_feature_spec.rb:8

-------------------------------------------------------------------------------------------------------------------------------
arel

### DIFF
--- a/app/models/clickhouse_dictionary/base.rb
+++ b/app/models/clickhouse_dictionary/base.rb
@@ -26,7 +26,7 @@ module ClickhouseDictionary
 
     def call
       fields = attribute_names
-      rows = scoped_collection.pluck(*pluck_fields).map { |values| fields.zip(values).to_h }
+      rows = scoped_collection.pluck(Arel.sql(pluck_fields.join(', '))).map { |values| fields.zip(values).to_h }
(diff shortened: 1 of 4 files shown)


app/models/clickhouse_dictionary/base.rb
```diff
@@ -26,7 +26,7 @@ def call
 
     def call
       fields = attribute_names
-      rows = scoped_collection.pluck(*pluck_fields).map { |values| fields.zip(values).to_h }
+      rows = scoped_collection.pluck(Arel.sql(*pluck_fields.join(', '))).map { |values| fields.zip(values).to_h }
```
gigorok, line 29: remove splat operator here

Ivanov-Anton:
This change solves the issue number #570

PR #698, opened 2020-05-08
raise error when we get deprecation warning in specs

Ivanov-Anton:
This change will continue to check the code for warnings like DEPRECATION WARNING and generate
`ActiveSupport::DeprecationException` exception.

In the future, when an error of this type occurs, the programmer is required to solve this problem.


gigorok:
@Ivanov-Anton rebase please on top of master branch to see rest of deprecations if exists

Issue #596, opened 2020-01-16
Outgoing registration fields


dmitry-sinina:
additional fields for outgoing registration:

```
last_reply_contacts
last_request_contact
node_id
pop_id
```
JRPC response example:
```
{
    "id": "ZvNxbLHCKPJxYiPrEZ1MJ1SDilWOhrX3",
    "jsonrpc": "2.0",
    "result": [
        {
            "attempt": 0,
            "auth_user": "1234567",
            "contact": "sip:1234567@1.1.1.1:5060",
            "display_name": "",
            "domain": "example.com",
            "event_sink": "",
            "expires": 1800,
            "expires_interval": 0,
            "expires_left": 1277,
            "force_reregister": false,
            "handle": "14-33A5276E-5EC072590004BF6E-5DC577C0",
            "id": "10",
            "last_error_code": 0,
            "last_error_initiator": null,
            "last_error_reason": null,
            "last_reply_contacts": "sip:1234567@1.1.1.1:5060;transport=udp",
            "last_request_contact": "sip:1234567@1.1.1.1:5060",
            "last_request_time": 1589705628,
            "last_succ_reg_time": 1589705628,
            "max_attempts": 0,
            "node_id": 14,
            "pop_id": 3,
            "postpone_timeout_msec": 0,
            "proxy": "",
            "proxy_transport_protocol_id": 1,
            "retry_delay": 5,
            "state": "RegisterActive",
            "state_code": 1,
            "transport_protocol_id": 1,
            "user": "1234567"
        }
    ]
}
```

senid231:
model class `RealtimeData::OutgoingRegistration`
query object `Yeti::OutgoingRegistrations#search`

attributes should be added to model class and to test stubs

need to add feature tests for checking that new fields are shown in admin UI

senid231:
@dmitry-sinina solved

PR #705, opened 2020-05-14
added outgoing registration fields

Ivanov-Anton:
additional fields for outgoing registration:
`

- last_reply_contacts
- last_request_contact
- pop_id
`
tring to solve issue https://github.com/yeti-switch/yeti-web/issues/596

### DIFF
--- a/spec/features/realtime_data/outgoing_resgistrations/index_outgoing_registrations_spec.rb
+++ b/spec/features/realtime_data/outgoing_resgistrations/index_outgoing_registrations_spec.rb
@@ -6,7 +6,10 @@ describe 'Index Outgoing Registrations', type: :feature do
   include_context :login_as_admin
 
   let!(:node) { FactoryBot.create(:node) }
-  let(:record_attributes) { FactoryBot.attributes_for(:outgoing_registration, :filled) }
+  let!(:pop) { FactoryBot.create(:pop) }
+  let(:record_attributes) do
+    FactoryBot.attributes_for(:outgoing_registration, :filled, pop_id: pop.id, node_id: node.id)
+  end
   before do
     stub_jrpc_request('show.registrations', node.rpc_endpoint).and_return([record_attributes.stringify_keys])
     visit outgoing_registrations_path(q: { node_id_eq: node.id })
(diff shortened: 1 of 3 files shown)


spec/features/realtime_data/outgoing_resgistrations/index_outgoing_registrations_spec.rb
```diff
@@ -6,7 +6,9 @@
   include_context :login_as_admin
 
   let!(:node) { FactoryGirl.create(:node) }
+  let!(:pop) { FactoryGirl.create(:pop) }
   let(:record_attributes) { FactoryGirl.attributes_for(:outgoing_registration, :filled) }
```
senid231, line 10: add new attributes to `factory :outgoing_registration`

spec/features/realtime_data/outgoing_resgistrations/index_outgoing_registrations_spec.rb
```diff
@@ -17,4 +19,18 @@
...
+      visit outgoing_registrations_path(q: { node_id_eq: node.id })
+    end
+
+    it 'has record' do
+      expect(page).to have_css('.col-id', text: record_filled[:id])
+      expect(page).to have_css('.col-last_reply_contacts')
+      expect(page).to have_css('.col-last_request_contact')
+      expect(page).to have_css('.col-pop', text: pop.name)
+    end
+  end
```
senid231, line 35: you don't need to add `before` in this context.
`before` from parent `context/describe` will be executed in nested context

in this context you need to override `record_attributes` and check that additional fields are shown correctly in table

like this
```ruby
  context 'with additional fields' do
    let(:record_attributes) do
      attributes_for(:outgoing_registration, :filled, pop: pop, last_reply_contacts: 'foo', last_request_contact: 'bar')
    end

    it 'has record' do
      expect(page).to have_css('.col-id', text: record_attributes[:id])
      expect(page).to have_css('.col-last_reply_contacts', text: record_attributes[:last_reply_contacts])
      expect(page).to have_css('.col-last_request_contact', text: record_attributes[:last_request_contact])
      expect(page).to have_css('.col-pop', text: pop.name)
    end
  end
```

Ivanov-Anton: Should I check the value of the added fields?

senid231: yes

Ivanov-Anton: Can't fount  pop in table and in record_attributes Hash missing also ()
part from output error rspec:
`but there were no matches. Also found "pop", ""`
byebug:
record_attributes.key?(:pop_id) #=> false

Ivanov-Anton:
```
def initialize(nodes, pops, params = {})
  @params = clean_search_params(params).with_indifferent_access
  @errors = []
  set_nodes(nodes)
  set_pops(pops)
end
  
```

Ivanov-Anton: Maybe I should add parameter for initialize OutgoingRegistrations

senid231:
CI passes
https://travis-ci.org/github/yeti-switch/yeti-web/builds/689769280

but here it shows like pending. Looks like bug on github

PR #715, opened 2020-05-19
upgrade rails to 5.2.4.3 (fix vulnerabilities)

Ivanov-Anton:


Gemfile
```diff
@@ -7,7 +7,7 @@ gem 'activemodel-serializers-xml'
 gem 'pg'
 gem 'postgres_ext', git: 'https://github.com/didww/postgres_ext.git', branch: 'rails-5-2'
 gem 'rack', '2.1.3' # https://github.com/rack/rack/issues/1628
-gem 'rails', '~> 5.2.4'
+gem 'rails', '~> 5.2.4.2'
```
gigorok, line 10: this change is not needed

gigorok: running bundle update rails will be enough

Issue #716, opened 2020-05-20
webdrivers gem instead of deprecated chromedriver-helper


gigorok:

gigorok:
https://github.com/titusfortner/webdrivers

PR #722, opened 2020-05-28
replace gem provided chromedriver

Ivanov-Anton:
solved: #716 

[**Webdrivers gem**](https://github.com/titusfortner/webdrivers) will check the latest supported driver and automatically download it unless you set a specific version.
This gem automatically will be loaded via Railtie

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -107,10 +107,10 @@ gem 'apitome', '~> 0.1.0'
 group :test do
   gem 'capybara'
   gem 'capybara-screenshot'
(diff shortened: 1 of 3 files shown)


Ivanov-Anton:
When added this gem I got error from webmock `WebMock::NetConnectNotAllowedError:`
To solve this error added whitelist the url:  `allow: "chromedriver.storage.googleapis.com"`
 so Webdrivers can download the driver.  [link issue about this case](https://github.com/titusfortner/webdrivers/issues/4)

PR #744, opened 2020-06-17
resolve #743, fixed filters named routing_tag_ids_array_contains and tagged 

### LINKED ISSUE #743: Routing Dialpeers Filter Issues
2 Filter issues in Routing -> Dialpeers section
1. **ROUTING TAG IDS CONTAINS** filter is not working
Result: We're sorry, but something went wrong.

2. **Tagged filter = No** option is not working
Result: filters the same dialpeers as it is with option Yes
It should filter dialpeers without tagged assigned == NOT TAGGED

Ivanov-Anton:
Summary: filters called routing tag ids contains, tagger do not work

1. To correct the error's filter routing_tag_ids_array_contains was implemented differently, with the scope and method `contains` from `active_record_extended` gem
1. To fix tagged filter was removed Array arguments (from *value to vaue) because:
`ActiveModel::Type::Boolean.new.cast(['true'])` => true
`ActiveModel::Type::Boolean.new.cast(['false'])` => true too

**tests**

**app/models/customers_auth.rb**
    src_prefix
    dst_prefix
    uri_domain
    from_domain
    to_domain
    x_yeti_auth
**app/models/dialpeer.rb**
routing_tag_ids_array_contains
tegged
**app/models/routing/destination.rb**
    routing_tag_ids_array_contains
    tagged
**app/models/routing/routing_tag_detection_rule.rb**
    routing_tag_ids_array_contains

### DIFF
--- a/app/models/concerns/routing_tag_ids_scopeable.rb
+++ b/app/models/concerns/routing_tag_ids_scopeable.rb
@@ -8,7 +8,7 @@ module RoutingTagIdsScopeable
       where("yeti_ext.tag_compare(routing_tag_ids, ARRAY[#{id.join(',')}], routing_tag_mode_id)>0")
     }
 
-    scope :tagged, lambda { |*value|
+    scope :tagged, lambda { |value|
       if ActiveModel::Type::Boolean.new.cast(value)
         where("routing_tag_ids <> '{}'") # has tags
       else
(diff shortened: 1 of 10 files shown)


gigorok:
https://github.com/yeti-switch/yeti-web/issues/743

Issue #336, opened 2018-07-13
When create a dialpeer via API, server doesn't check params


sashker:
At least, I've found two params which rise 500 Internal server error. I understand that the problem relates to data I send, but I believe, the server should check passed parameters.

For example:
Jul 13 03:53:23 yeti YETI-admin[16302]: [03:53:23.119290 ] [ERROR]: Internal Server Error: PG::NotNullViolation: ERROR:  null value in column "lcr_rate_multiplier" violates not-null constraint#012DETAIL:  Failing row contains (2309431, t, , , , 0, 0, null, 24, 0.0, 0.0, 70, 69, , , f, 0, 1, null, 0.0, 1, 1, 2018-07-13 03:53:22.907862+00, 2023-07-13 03:53:22.907863+00, 6, 0, null, 2018-07-13 03:53:23.077164+00, 0, null, null, null, null, f, 0, 100, f, {}).#012: INSERT INTO "dialpeers" ("valid_from", "valid_till", "enabled", "initial_rate", "next_rate", "prefix", "src_rewrite_rule", "src_rewrite_result", "dst_rewrite_rule", "dst_rewrite_result", "priority", "capacity", "lcr_rate_multiplier", "force_hit_rate", "short_calls_limit", "routing_group_id", "vendor_id", "account_id", "gateway_group_id", "created_at") VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19, $20) RETURNING "id" /home/yeti-web/vendor/bundle/ruby/2.3.0/gems/activerecord-4.2.9/lib/active_record/connection_adapters/postgresql_adapter.rb:602:in `exec_prepared'#012/home/yeti-web/vendor/bundle/ruby/2.3.0/gems/activerecord-4.2.9/lib/active_record/connection_adapters/postgresql_adapter.rb:602:in `block in

Also, something similar happens if I pass gateway_id = 0.

gigorok:
add required validators to Dialpeer model

Ivanov-Anton:
Added missing presence validations for field: lcr_rate_multiplier and gateway_id with zero

PR #747, opened 2020-06-25
fix bug: missing validates for dialpeers model class

Ivanov-Anton:
refs #336 

Added missing presence validations for field: **lcr_rate_multiplier** and **gateway_id** with zero. 

we have error when pass gateway_id with 0 because in throws out error `undefined method vendor_id for nil class` in [200](https://github.com/yeti-switch/yeti-web/blob/4d0aad6a414d2b8b8520f66faccce590f7b06cdb/app/models/dialpeer.rb#L200) line

### DIFF
--- a/app/models/dialpeer.rb
+++ b/app/models/dialpeer.rb
@@ -67,8 +67,8 @@ class Dialpeer < Yeti::ActiveRecord
   validates :account, :routing_group, :vendor, :valid_from, :valid_till,
                         :initial_rate, :next_rate,
                         :initial_interval, :next_interval, :connect_fee,
-                        :routing_tag_mode, :routeset_discriminator, presence: true
-  validates :initial_rate, :next_rate, :connect_fee, numericality: true
+                        :routing_tag_mode, :routeset_discriminator, :lcr_rate_multiplier, presence: true
+  validates :initial_rate, :next_rate, :connect_fee, :lcr_rate_multiplier, numericality: true
   validates :initial_interval, :next_interval, numericality: { greater_than: 0 } # we have DB constraints for this
   validates :acd_limit, numericality: { greater_than_or_equal_to: 0.00 }
   validates :asr_limit, numericality: { greater_than_or_equal_to: 0.00, less_than_or_equal_to: 1.00 }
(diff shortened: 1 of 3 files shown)


Ivanov-Anton:
additional checks added:
only_integer: true for  next_interval and initial_interval because both defined as integer

spec/requests/api/rest/admin/dialpeer_spec.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+RSpec.describe Api::Rest::Admin::DialpeersController, type: :request do
+  include_context :json_api_admin_helpers, type: :dialpeers
+  include_context :init_routing_tag_collection
+  describe 'POST /api/rest/admin/dialpeers without gateway_id and gateway_group_id' do
```
gigorok, line 6: instead of 
describe 'POST /api/rest/admin/dialpeers without gateway_id and gateway_group_id' do
use inner block
describe 'POST /api/rest/admin/dialpeers' do
  context 'without gateway_id and gateway_group_id' do
    ...
  end
  ...
end

app/models/dialpeer.rb
```diff
@@ -181,23 +181,23 @@ def contractor_is_vendor
...
 
-    unless gateway.allow_termination
+    unless gateway&.allow_termination
       errors.add(:gateway, 'must be allowed for termination')
     end
   end
 
   def vendor_owners_the_gateway_group
-    errors.add(:gateway_group, 'must be owned by selected vendor') unless gateway_group_id.nil? || (vendor_id && gateway_group_id && vendor_id == gateway_group.vendor_id)
+    errors.add(:gateway_group, 'must be owned by selected vendor') unless gateway_group_id.nil? || (vendor_id && gateway_group_id && vendor_id == gateway_group&.vendor_id)
```
gigorok, line 200: validate :vendor_owners_the_gateway_group, if: :gateway_group

app/models/dialpeer.rb
```diff
@@ -181,23 +181,23 @@ def contractor_is_vendor
...
-    return true if gateway_id.nil? || gateway.is_shared?
+    return true if gateway_id.nil? || gateway&.is_shared?
 
-    unless vendor_id && vendor_id == gateway.contractor_id
+    unless vendor_id && vendor_id == gateway&.contractor_id
       errors.add(:gateway, 'must be owned by selected vendor or be shared')
     end
 
-    unless gateway.allow_termination
+    unless gateway&.allow_termination
```
gigorok, line 194: validate :vendor_owners_the_gateway, if: :gateway instead

app/models/dialpeer.rb
```diff
@@ -181,23 +181,23 @@ def contractor_is_vendor
   end
 
   def vendor_owners_the_account
-    errors.add(:account, 'must be owned by selected vendor') unless account_id.nil? || (vendor_id && vendor_id == account.contractor_id)
+    errors.add(:account, 'must be owned by selected vendor') unless account_id.nil? || (vendor_id && vendor_id == account&.contractor_id)
```
gigorok, line 184: validate :vendor_owners_the_account, if: :account

Issue #334, opened 2018-07-08
Audit log issue after moving Destination to Routing::Destination


dmitry-sinina:
```
Jul  8 03:47:33 yeti-demo YETI-admin[27242]: [03:47:33.539575 ] [ INFO]: Started GET "/audit_log_items?order=id_desc&page=7" for 127.0.0.1 at 2018-07-08 03:47:33 +0300
Jul  8 03:47:33 yeti-demo YETI-admin[27242]: [03:47:33.540892 ] [ INFO]: Processing by AuditLogItemsController#index as HTML
Jul  8 03:47:33 yeti-demo YETI-admin[27242]: [03:47:33.540974 ] [ INFO]: Parameters: {"order"=>"id_desc", "page"=>"7"}
Jul  8 03:47:33 yeti-demo YETI-admin[27242]: [03:47:33.552868 ] [ INFO]: Rendering vendor/bundle/ruby/2.3.0/gems/activeadmin-1.0.0/app/views/active_admin/resource/index.html.arb
Jul  8 03:47:33 yeti-demo YETI-admin[27242]: [03:47:33.696704 ] [ INFO]: Rendered vendor/bundle/ruby/2.3.0/gems/activeadmin-1.0.0/app/views/active_admin/resource/index.html.arb (143.1ms)
Jul  8 03:47:33 yeti-demo YETI-admin[27242]: [03:47:33.697569 ] [ERROR]: uninitialized constant Destination#012Did you mean?  DestinationDecorator
```

https://github.com/yeti-switch/yeti-web/commit/884bc6ed5780f52fc9eed37b0b4c22688162263c

We shoud update audit log table and rename **Destination** to **Routing::Destination**

gigorok:
we have to create a migration like this
UPDATE audit_log_items SET item_type = 'Routing::Destination' WHERE item_type = 'Destination'

PR #751, opened 2020-06-29
refs #334 rename audit log items than contains old class name "Destination" to "Routing::Destination"

Ivanov-Anton:
Renamed items than contains old class name "Destination" in items_type column


db/migrate/20200630083409_rename_items_from_gui_versions_table.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class RenameItemsFromGuiVersionsTable < ActiveRecord::Migration[5.2]
+  def change
+    suppress_messages do
```
gigorok, line 5: do not use suppress_messages in migrations

PR #762, opened 2020-07-06
 enable eager_load in test env

### LINKED ISSUE #761: enable eager_load in test env
to prevent situation with non covered by specs broken code to go to the master branch we can enable loading whole code while booting application. 
FYI: it will increase a bit starting application time in test environment

Ivanov-Anton:
to prevent situation with non covered by specs broken code to go to the master branch we can enable loading whole code while booting application

closes #761


spec/routing/route_spec.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'lib/call_sql/yeti'
+
```
gigorok, line 4: rollback this changes

PR #780, opened 2020-09-08
Dialpeer import edit id (valid till)

sowecastellano:
Hi,

Can anyone please help on importing dialpeers?

We try exporting .csv of current dialpeer, editing valid dates of dialpeer and importing + update but yeti-web creates a new id instead of editing current dialpeer.

Captura de pantalla 2020-09-08 a las 15 57 15


dmitry-sinina:
@sowecastellano  https://community.yeti-switch.org/

PR #781, opened 2020-09-19
update readme about project setup

senid231:


dmitry-sinina:
postgresql-12 should be used with master branch

PR #790, opened 2020-11-14
invoice creation without update cdrs.vendor_invoice_id

senid231:
it will speed up invoice generation dramatically


senid231:
@dmitry-sinina floating failed test

can be merged

PR #829, opened 2021-01-14
Admin api gateway attributes

### LINKED ISSUE #821: Admin API gateway attributes
**incoming-auth-username** and **incoming-auth-password** attributes should be available in gateway resource.

paratskiy:
fixes #821 
* [x] add create/update/show tests for current gateway resource
* [x] Write tests for new gateway attributes
* [x] Add new attributes for pass all tests
* [x] Change API documentation

### DIFF
--- a/spec/requests/api/rest/admin/gateways_spec.rb
+++ b/spec/requests/api/rest/admin/gateways_spec.rb
@@ -3,73 +3,7 @@
 RSpec.describe Api::Rest::Admin::GatewaysController, type: :request do
   include_context :json_api_admin_helpers, type: :gateways
 
-  let!(:gateway) { FactoryBot.build(:gateway) }
   # after { puts pretty_response_json }
-
-  let!(:gateway_response_attributes) do
-    {
-      'name': gateway.name,
-      'enabled': gateway.enabled,
-      'priority': gateway.priority,
-      'weight': gateway.weight,
-      'acd-limit': gateway.acd_limit,
-      'asr-limit': gateway.asr_limit,
-      'allow-origination': gateway.allow_origination,
-      'allow-termination': gateway.allow_termination,
-      'sst-enabled': gateway.sst_enabled,
-      'host': gateway.host,
-      'port': gateway.port,
-      'resolve-ruri': gateway.resolve_ruri,
-      'diversion-rewrite-rule': gateway.diversion_rewrite_rule,
-      'diversion-rewrite-result': gateway.diversion_rewrite_result,
-      'src-name-rewrite-rule': gateway.src_name_rewrite_rule,
-      'src-name-rewrite-result': gateway.src_name_rewrite_result,
-      'src-rewrite-rule': gateway.src_rewrite_rule,
-      'src-rewrite-result': gateway.src_rewrite_result,
-      'dst-rewrite-rule': gateway.dst_rewrite_rule,
-      'dst-rewrite-result': gateway.dst_rewrite_result,
-      'auth-enabled': gateway.auth_enabled,
-      'auth-user': gateway.auth_user,
-      'auth-password': gateway.auth_password,
-      'auth-from-user': gateway.auth_from_user,
-      'auth-from-domain': gateway.auth_from_domain,
-      'term-use-outbound-proxy': gateway.term_use_outbound_proxy,
-      'term-force-outbound-proxy': gateway.term_force_outbound_proxy,
-      'term-outbound-proxy': gateway.term_outbound_proxy,
-      'term-next-hop-for-replies': gateway.term_next_hop_for_replies,
-      'term-next-hop': gateway.term_next_hop,
-      'term-append-headers-req': gateway.term_append_headers_req,
-      'sdp-alines-filter-list': gateway.sdp_alines_filter_list,
-      'ringing-timeout': gateway.ringing_timeout,
-      'relay-options': gateway.relay_options,
-      'relay-reinvite': gateway.relay_reinvite,
-      'relay-hold': gateway.relay_hold,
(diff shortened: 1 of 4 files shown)


spec/requests/api/rest/admin/gateways_spec.rb
```diff
@@ -433,4 +526,101 @@
...
+      {
+        'name': gateway.name,
+        'enabled': gateway.enabled,
+        'priority': gateway.priority,
+        'weight': gateway.weight,
+        'acd-limit': gateway.acd_limit,
+        'asr-limit': gateway.asr_limit,
+        'host': gateway.host
+      }
+    end
```
senid231, line 582: use actual valus in hash, like
```ruby
    let(:json_api_request_attributes) do
      {
        name: 'test gateway',
        enabled: false,
        # ....
      }
    end
```

also no need to write symbol in single quote if it has only letters/digits

so `{ 'host': 'test.com' }` and `:'host'` should be written like  `{ host: 'test.com' }` and `:host`

but `{ 'asr-limit': 123 }` and `:'asr-limit'` can be written only in quotes

senid231: same for other hashes in this file

spec/requests/api/rest/admin/gateways_spec.rb
```diff
@@ -433,4 +526,101 @@
...
+
+    let(:last_gateway) { Gateway.last! }
+
+    include_examples :returns_json_api_record, relationships: gateway_relationship_names, status: 201 do
+      let(:json_api_record_id) { last_gateway.id.to_s }
+      # Without all attributes test 'expect(actual_relationships).to match_array(relationships)' was falid
+      let(:json_api_record_attributes) { gateway_response_attributes }
+    end
+
+    include_examples :changes_records_qty_of, Gateway, by: 1
```
senid231, line 625: I think we could do better testing by replacing

```ruby
include_examples :changes_records_qty_of, Gateway, by: 1
```

with more extended test

```ruby
it 'creates gateway with correct attributes' do
  expect { subject }.to change { Gateway.count }.by(1)

  expect(last_gateway).to have_attributes(
                                           name: json_api_request_attributes[:name],
                                           # ....
                                           contractor: contractor,
                                           # ...
                                         )
end
```

senid231 (changes requested):

spec/requests/api/rest/admin/gateways_spec.rb
```diff
@@ -3,6 +3,33 @@
 RSpec.describe Api::Rest::Admin::GatewaysController, type: :request do
   include_context :json_api_admin_helpers, type: :gateways
 
+  after { puts pretty_response_json }
```
senid231, line 6: remove this line

it is useful when creating test but should be removed before merge (to not spam into stdout)

spec/requests/api/rest/admin/gateways_spec.rb
```diff
@@ -433,4 +460,236 @@
...
+    let(:last_gateway) { Gateway.last! }
+
+    include_examples :returns_json_api_record, relationships: gateway_relationship_names, status: 201 do
+      let(:json_api_record_id) { last_gateway.id.to_s }
+      let(:json_api_record_attributes) do
+        hash_including(json_api_request_attributes)
+      end
+    end
+
+    # include_examples :changes_records_qty_of, Gateway, by: 1
```
senid231, line 626: remove this comment

spec/requests/api/rest/admin/gateways_spec.rb
```diff
@@ -433,4 +460,236 @@
...
+      {
+        name: 'Test name',
+        enabled: false,
+        priority: 100,
+        weight: 100,
+        'acd-limit': 0.0,
+        'asr-limit': 0.0,
+        host: 'test.example.com',
+        'incoming-auth-username': 'incoming_username',
+        'incoming-auth-password': 'incoming_password'
```
senid231, line 581: do not add new attributes here

youre POST describe should look like this

```ruby
  describe 'POST /api/rest/admin/gateways' do
    # ...

    let(:json_api_request_attributes) do
      # only required attributes
    end

    include_examples :returns_json_api_record, relationships: gateway_relationship_names, status: 201 do
      let(:json_api_record_id) { last_gateway.id.to_s }
      let(:json_api_record_attributes) do
        hash_including(json_api_request_attributes)
      end
    end

    it 'creates gateway with correct attributes' do
      expect { subject }.to change { Gateway.count }.by(1)

      expect(last_gateway).to have_attributes(
        # ...
      )
    end

    context 'with incoming-auth-username and incoming-auth-password' do
      let(:json_api_request_attributes) do
  	super().merge 'incoming-auth-username': 'incoming_username',
                      'incoming-auth-password': 'incoming_password'
      end

      include_examples :returns_json_api_record, relationships: gateway_relationship_names, status: 201 do
        let(:json_api_record_id) { last_gateway.id.to_s }
        let(:json_api_record_attributes) do
          hash_including(json_api_request_attributes)
        end
      end

      it 'creates gateway with correct attributes' do
        expect { subject }.to change { Gateway.count }.by(1)

        expect(last_gateway).to have_attributes(
          # ...
        )
      end
    end

  end
```

senid231: because we need to keep test where only required attributes are passed - to ensure that you can create gateway without passing new attributes

senid231: we will not add new context on PATCH describe because in case of PATCH we need only to test that attributes are updatable

senid231 (changes requested):

senid231 (approved):

Issue #828, opened 2021-01-13
Autogenerate gateway credentials


dmitry-sinina:
Allow to autogenerate **Incoming auth username** and **Incoming auth password**  for **Gateway** admin UI

rules:
* length = 20
* allowed symbols = a-z,A-Z, 0-9

Also we need autogeneration for username/password at **System->API Accesses**.

In both cases customer should be able to enter his own values.

- by default input fields should be empty - no auto population required
- admin should be able to click on hint for Incoming auth username (<strong>click<strong> to fill random username)
- admin should be able to click on hint for Incoming auth password (<strong>click<strong> to fill random password)

PR #838, opened 2021-01-20
Autogenerate gateway and api accesses credentials

Skumring:
Fixes https://github.com/yeti-switch/yeti-web/issues/828


dmitry-sinina:
would not be better to implement it on browser side, on js?

dmitry-sinina:
also looks like current implementation will fill **incoming_auth_username/password** with random data on every gw creation. It is not ok because it will force customer to create gateways with username/password when this parameter is optional.

Skumring:
@dmitry-sinina implemented generator via JS and used it for Gateway forms - will generate the new credentials by the double click on the input field (added hints below input fields).
new_credential_form_credential_generator

Skumring:
From now generator running only manually by the click on the link in a hint of the input field.
new_gateway_form
new_api_accesses_form

PR #840, opened 2021-01-21
Additional filters in cdr

### LINKED ISSUE #801: Feature request: Additional Filters ROUTING TAGs in CDR History
It would be nice to include **ROUTING TAG filters** in **CDR History** section, same filters are already available in Routing -> Diapeers section.

Filters:

ROUTING TAG IDS COVERS
ROUTING TAG IDS CONTAINS
TAGGED: Yes/No

Thank you.

paratskiy:
fixes #801

* [x] Add ROUTING TAG IDS CONTAINS, TAGGED: Yes/No filters to cdr history
* [x] Add tests for additional filters
* [x] Rewrite feature tests for filtering by Tagged

### DIFF
--- a/app/models/cdr/cdr.rb
+++ b/app/models/cdr/cdr.rb
@@ -162,7 +162,6 @@ class Cdr::Cdr < Cdr::Base
   ].freeze
 
   include Partitionable
-  include RoutingTagIdsScopeable
   self.pg_partition_name = 'PgPartition::Cdr'
 
   belongs_to :rateplan, class_name: 'Routing::Rateplan', foreign_key: :rateplan_id
@@ -322,6 +321,14 @@ class Cdr::Cdr < Cdr::Base
 
   scope :routing_tag_ids_array_contains, ->(*tag_id) { where.contains routing_tag_ids: Array(tag_id) }
 
+  scope :tagged, lambda { |value|
+    if ActiveModel::Type::Boolean.new.cast(value)
+      where("routing_tag_ids <> '{}'") # has tags
+    else
+      where("routing_tag_ids = '{}'") # no tags
+    end
+  }
+
   private
 
   def self.ransackable_scopes(_auth_object = nil)
@@ -331,7 +338,6 @@ class Cdr::Cdr < Cdr::Base
       account_id_eq
       routing_tag_ids_include
       routing_tag_ids_array_contains
-      routing_tag_ids_covers
       tagged
     ]
   end
--- a/app/models/cdr/cdr.rb
+++ b/app/models/cdr/cdr.rb
@@ -162,6 +162,7 @@ class Cdr::Cdr < Cdr::Base
   ].freeze
 
   include Partitionable
+  include RoutingTagIdsScopeable
   self.pg_partition_name = 'PgPartition::Cdr'
 
   belongs_to :rateplan, class_name: 'Routing::Rateplan', foreign_key: :rateplan_id
@@ -319,6 +320,8 @@ class Cdr::Cdr < Cdr::Base
     SqlCaller::Cdr.select_value("select pending_events from pgq.get_consumer_info('cdr_billing', 'cdr_billing')")
   end
 
+  scope :routing_tag_ids_array_contains, ->(*tag_id) { where.contains routing_tag_ids: Array(tag_id) }
+
   private
 
   def self.ransackable_scopes(_auth_object = nil)
@@ -327,6 +330,9 @@ class Cdr::Cdr < Cdr::Base
       status_eq
       account_id_eq
       routing_tag_ids_include
+      routing_tag_ids_array_contains
+      routing_tag_ids_covers
+      tagged
     ]
   end
 end
--- a/app/models/cdr/cdr.rb
+++ b/app/models/cdr/cdr.rb
@@ -325,7 +325,7 @@ class Cdr::Cdr < Cdr::Base
     if ActiveModel::Type::Boolean.new.cast(value)
       where("routing_tag_ids <> '{}'") # has tags
     else
-      where("routing_tag_ids = '{}'") # no tags
+      where("routing_tag_ids = '{}' OR routing_tag_ids IS NULL") # no tags
     end
   }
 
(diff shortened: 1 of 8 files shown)


app/models/cdr/cdr.rb
```diff
@@ -162,6 +162,7 @@ class Cdr::Cdr < Cdr::Base
   ].freeze
 
   include Partitionable
+  include RoutingTagIdsScopeable
```
senid231, line 165: you can copy `scope :tagged, lambda { |value| ... }` here instead of including module because second scope is not needed

spec/features/cdr/cdr_history/cdr_filter_spec.rb
```diff
@@ -0,0 +1,48 @@
+RSpec.describe 'Cdrs index page', type: :feature, js: true do
```
senid231, line 1: when spec file is in `features` folder you don't need to set `type: :feature` explicitly

rspec config `config.infer_spec_type_from_file_location!` will do it for you

spec/features/cdr/cdr_history/cdr_filter_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  include_context :login_as_admin
+  let!(:cdrs) { create_list(:cdr, 2) }
+
+  before do
+    Cdr::Cdr.delete_all
+    visit cdrs_path
+  end
+
+  describe 'filtering', js: true do
```
senid231, line 14: you can squash it with upper describe and remove one level of block nesting

```ruby
RSpec.describe 'Cdrs index page filtering', js: true do
  # ...
end
```

also parent context/describe meta options are propagated to inner contexts and tests, so no need for inside context/describe which already has it `js: true`

spec/features/cdr/cdr_history/cdr_filter_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      #
+      # end
+
+      let!(:tag) { create :routing_tag }
+      let!(:cdr_tagged) { create :cdr, routing_tag_ids: [tag.id] }
+
+      it 'shows filtered records with routing_tags' do
+        within_filters do
+          select :Yes, from: :Tagged
+        end
```
senid231, line 26: it would be nicer if you do filter selection in `before do ... end` and provide value as `let()`

```ruby
before do
   within_filters do
    fill_in_chosen 'Tagged', with: tagged_filter_value
  end
end
let(:tagged_filter_value) { 'Yes' }
```

also what is `select :Yes, from: :Tagged` ? is it working? I'm not familiar with that

senid231 (changes requested):

spec/features/cdr/cdr_history/cdr_filter_spec.rb
paratskiy: select :Yes, from: :Tagged it workin, saw in spec/features/routing/dialpeers/filters_spec.rb. But i will rewrite this

paratskiy: fill_in_chosen method is not suitable for this check, used method select_by_value

spec/features/cdr/cdr_history/cdr_filter_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    it 'shows filtered records with contains ids' do
+      within_filters do
+        fill_in_chosen 'Routing Tag IDs Contains', with: tagged_filter_value
+      end
+
+      subject
+      expect(page).to have_table
+      expect(page).to have_table_row count: 1
+      expect(page).to have_table_cell column: 'Id', text: cdr_tagged.id
+      expect(page).to have_select 'q_routing_tag_ids_array_contains', selected: tagged_filter_value
```
senid231, line 69: we should not use `have_select` here because in UI we don't show ordinary select - it being hidden by chosen.js which shows custom built element. So this test can be false positive - appropriate value can be selected in `<select>` but chosen container will be empty.

So you need to replace expectation with following

```ruby
expect(page).to have_field_chosen('Routing Tag IDs Contains', with: tagged_filter_value)
```

and add following matcher to `Helpers::Chosen`

```ruby
  # Checks chosen select presence on page
  # @param label [String] field label.
  # @param options [Hash]
  #   :with [String, nil] check selected option text whether not nil,
  #   :disabled [Boolean, nil] default false,
  #   :exact [Boolean] match :with by exact text (default true),
  #   for other options @see #have_selector.
  def have_field_chosen(label, options = {})
    with = options.delete(:with)
    disabled = options.delete(:disabled)
    exact = options.delete(:exact)
    exact = true if exact.nil?
    warn 'empty :with will be ignored because :exact is false' if with.blank? && !exact
    options[exact ? :exact_text : :text] = with unless with.nil?
    selector = chosen_container_selector(label, disabled: disabled)
    have_selector(selector, options)
  end
```

senid231: same for `Tagged`

spec/features/cdr/cdr_history/cdr_filter_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  context 'with filter by not tagged' do
+    let(:tagged_filter_value) { false }
+
+    it 'shows filtered records without routing_tags' do
+      subject
+      expect(page).to have_table
+      expect(page).to have_table_row count: 2
+      expect(page).to have_table_cell column: 'Id', text: cdrs.first.id
```
senid231, line 41: this test does not check that tagged cdr does not appear on page. 

to test it you could add just

```ruby
expect(page).to_not have_table_cell column: 'Id', text: cdr_tagged.id
```
but I suggest another approach: we can check that only correct records are shown:

```ruby
expect(page).to have_table_row count: cdrs.size
cdrs.each { |cdr| expect(page).to have_table_cell column: 'Id', text: cdr.id }
```

this method is more readable and stable

senid231 (changes requested):

spec/features/cdr/cdr_history/cdr_filter_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+    context 'with filter by tagged' do
+      let(:filter_value) { 'Yes' }
+
+      it 'shows filtered records with routing_tags' do
+        subject
+        expect(page).to have_table
+        expect(page).to have_table_row count: cdrs_tagged.size
+        cdrs_tagged.each { |cdr| expect(page).to have_table_cell column: 'Id', text: cdr.id }
+        expect(page).to have_field_chosen('Tagged', with: 'Yes')
```
senid231, line 48: wrap this line in `within_filters do .. end`

it will be easier to understand that you check filter value here, not just value is some form

senid231: same for below tests

senid231 (changes requested):

senid231 (approved):

Issue #528, opened 2019-09-03
Additional UI filters


dmitry-sinina:
contractors:
- smtp connection(chosen)
- description
- address
- phones

payments:
- account(chosen)
- notes

invoices:
- amount
- calls count
- calls duration
- billing duration

gateways:
-   auth enabled
-   username
-   password
-   incoming auth username
-   incoming auth password
-   codec group

registrations:
- transport protocol
- domain
- username
- auth user
- auth password
- contact

customer auths:
- Require Incoming Auth
- incoming auth username
- incoming auth password
- check account balance

rateplans:
- profit control mode

destinations:
- valid from
- valid till
- Quick filter "Time valid"
- rate policy
- reverse billing
- initial interval
- next interval
- asr limit
- acd limit
- short calls limit

routing plans
- rate delta max
- max rerouting attempts
- routing groups

Lnp Routing Plan Lnp Rules:
- routing plan
- dst prefix
- database
- created at

numberlists:
- external id
- tag action

numberlist items:
- action
- tag action
- created at
- updated at

area prefixes:
- prefix covers

auth logs:
- code
- reason
- internal reason
- realm
- request method
- X-yeti-auth
- Diversion
- pai
- ppi
- privacy
- rpid
- rpid privacy

cdr history:

- lega remote ip
- lega local ip
- legb local ip
- legb remote ip
- legA originator ip
- uuid

cdr_exports:
- id
- rows count

PR #842, opened 2021-01-21
Added additional UI filters

Skumring:
Fixes https://github.com/yeti-switch/yeti-web/issues/528

#### Gateways
- `username` - attribute is missing in the `Gateway` model
- `password` - attribute is missing in the `Gateway` model
#### Customer Auths
- `incoming_auth_username` - attribute is missing in the `CustomersAuth` model, but the same attribute exists in the parent `Gateway`, do we need to filter customer auths by the `incoming_auth_username` from `Gateway` association, or we just no need this filter anymore?
- `incoming_auth_password` - attribute is missing in the `CustomersAuth` model, but the same attribute exists in the parent `Gateway`, do we need to filter customer auths by the `incoming_auth_password` from `Gateway` association, or we just no need this filter anymore?
#### Destinations
- `Quick filter "Time valid"` - not clear for me how that filter should work? It will be great to get more description or any examples.
#### Area Prefixes
- `prefix_covers` - attribute is missing in the `AreaPrefix` model, but it has a `prefix` attribute and this filter already exists.
#### Cdr History
- `lega_remote_ip` - attribute is missing in the `Cdr` model
- `lega_local_ip` - attribute is missing in the `Cdr` model
- `legb_local_ip` - attribute is missing in the `Cdr` model
- `legb_remote_ip` - attribute is missing in the `Cdr` model
- `lega_originator_ip` - attribute is missing in the `Cdr` model

### DIFF
--- a/app/admin/billing/contractors.rb
+++ b/app/admin/billing/contractors.rb
@@ -102,10 +102,14 @@ ActiveAdmin.register Contractor do
 
   filter :id
   filter :name
-  filter :enabled, as: :select, collection: [['Yes', true], ['No', false]]
-  filter :vendor, as: :select, collection: [['Yes', true], ['No', false]]
-  filter :customer, as: :select, collection: [['Yes', true], ['No', false]]
+  filter :address
+  filter :description
+  filter :phones
   filter :external_id
+  filter :smtp_connection, input_html: { class: 'chosen' }, collection: proc { System::SmtpConnection.pluck(:name, :id) }
+  boolean_filter :enabled
+  boolean_filter :vendor
+  boolean_filter :customer
 
   sidebar :links, only: %i[show edit] do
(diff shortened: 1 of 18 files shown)


Fivell (commented):
@Skumring take a look at another PR https://github.com/yeti-switch/yeti-web/pull/840/files

it includes specs as well

gigorok:
@dmitry-sinina could you explain what to do with non implemented filters ?

dmitry-sinina:
>username - attribute is missing in the Gateway model
>password - attribute is missing in the Gateway model

We need filters by **auth user**,**auth password**,**INCOMING AUTH USERNAME**, **INCOMING AUTH PASSWORD**

gigorok (approved):

Issue #678, opened 2020-04-22
Disabling audit log per model


dmitry-sinina:
Scope:
Currently next models can be audited via paper_trail
```
 Account, Billing::Contact, Billing::Invoice, CodecGroup, CodecGroupCodec, Contractor, CustomersAuth, Dialpeer, DialpeerNextRate, DisconnectCode, DisconnectPolicy, DisconnectPolicyCode, Equipment::Radius::AccountingProfile, Equipment::Radius::AccountingProfileInterimAttribute, Equipment::Radius::AccountingProfileStartAttribute, Equipment::Radius::AccountingProfileStopAttribute, Equipment::Radius::AuthProfile, Equipment::Radius::AuthProfileAttribute, Equipment::Registration, Equipment::SipOptionsProber, Gateway, GatewayGroup, GuiConfig, Node, Notification::Alert, Payment, Pop, Routing::Area, Routing::AreaPrefix, Routing::Destination, Routing::DestinationNextRate, Routing::Numberlist, Routing::NumberlistItem, Routing::RateGroup, Routing::Rateplan, Routing::RoutesetDiscriminator, Routing::RoutingPlan, Routing::RoutingPlanStaticRoute, Routing::RoutingTag, Routing::RoutingTagDetectionRule, RoutingGroup, System::ApiLogConfig, System::CdrConfig, System::LnpResolver, System::LoadBalancer, System::LuaScript, System::Network, System::NetworkPrefix, System::Sensor, System::SmtpConnection
```

We need to allow to disable auditing via config file.

```config/yeti_web.yml``` can have new attribute for instance
versioning_disable_for_models: [Account, Rateplan]` # yaml format

If model name is not present at config - paper trail should be enabled by default.
if model in array versioning_disable_for_models we should disable paper_trail for all application business logic.
 

Also if model is disabled for versioning , sidebar in admin ui should not be rendered
 
![image](https://user-images.githubusercontent.com/120269/105963475-b649ad80-6089-11eb-9069-82bc1311b3ac.png)

PR #849, opened 2021-01-26
Disabling audit log per model

Skumring:
Fixes https://github.com/yeti-switch/yeti-web/issues/678


lib/active_record/active_record_extension.rb
```diff
@@ -0,0 +1,28 @@
...
+    def inherited(subclass)
+      super
+
+      begin
+        # Paper trail should be enabled by default even if the class is not declared in the config.
+        # Object modification (actions :update, :touch) will not cause log saving if the class was declared with a 'false' value in the config.
+        options = {}
+        options[:class_name] = 'AuditLogItem' if defined?(AuditLogItem)
+        options[:on] = [:create, :destroy] if Rails.configuration.audit[subclass.name&.gsub('::', '/')] == false
+        subclass.send(:has_paper_trail, options) if %w[AuditLogItem PaperTrail::Version].exclude?(subclass.name)
```
gigorok, line 20: could we use paper_trail feature https://github.com/paper-trail-gem/paper_trail#per-class instead of monkey patching of activerecord? IMHO: this will be better

so leave paper_trail declaration in models as is, in a new initializer read audit.yml & disable specified models

lib/resource_dsl/acts_as_audit.rb
```diff
@@ -34,7 +34,7 @@ def show
       end
 
       action_item :history, only: %i[show edit] do
-        if authorized?(:history)
+        if authorized?(:history) && is_versioning_enabled_for_model?(resource)
```
gigorok, line 37: try to use `resource_class` instead, it will allow you to skip checking decoration
`is_versioning_enabled_for_model_class?(resource_class)`

config/yeti_web.yml.distr
```diff
@@ -28,3 +28,9 @@ sentry:
   dsn: "PUT_YOUR_DSN_HERE"
   node_name: "local"
   environment: "prod"
+
+versioning_disable_for_models: [
+  Routing/NumberlistItem,
```
gigorok, line 33: leave model class names as is (with ::)

config/initializers/paper_trail.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# Disabling creation of the new versions for each model declared under the key :versioning_disable_for_models in config/yeti_web.yml
+# Access to paper_trail methods like `object.versions` still present
+Rails.configuration.yeti_web['versioning_disable_for_models']&.each do |model_name|
+  normalized_model_name = model_name&.gsub('/', '::')&.safe_constantize
+  PaperTrail.request.disable_model(normalized_model_name) if normalized_model_name.present?
```
gigorok, line 7: do not check that `normalized_model_name` is present, when configuration has invalid model classes then the error should be raised. use constantize instead of safe_constantize

config/yeti_web.yml.distr
```diff
@@ -28,3 +28,9 @@ sentry:
...
+  Node,
+  Pop
```
gigorok, line 35: arrays in YAML can be described as
```
array:
  - item1
  - item2
```

spec/config/yeti_web_spec.rb
```diff
@@ -54,4 +55,42 @@
...
+    let(:vendor) { FactoryBot.create(:vendor) }
+    let(:account) { FactoryBot.create(:account, contractor: vendor) }
+    let(:numberlist_item) { FactoryBot.create(:numberlist_item) }
+    let(:node) { FactoryBot.create(:node) }
+
+    it 'paper trail should be enabled only for models non declared in the config' do
+      expect(PaperTrail.request.enabled_for_model?(vendor.class.name.safe_constantize)).to eq(true)
+      expect(PaperTrail.request.enabled_for_model?(account.class.name.safe_constantize)).to eq(true)
+      expect(PaperTrail.request.enabled_for_model?(numberlist_item.class.name.safe_constantize)).to eq(false)
+      expect(PaperTrail.request.enabled_for_model?(node.class.name.safe_constantize)).to eq(false)
```
gigorok, line 72: you already know classes, just use it instead of safe_constantize

lib/active_record/active_record_extension.rb
Fivell: @gigorok PaperTrail.request is designed to be used only per request since configuration stored in RequestStore

config/initializers/paper_trail.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+# Disabling creation of the new versions for each model declared under the key :versioning_disable_for_models in config/yeti_web.yml
+# Access to paper_trail methods like `object.versions` still present
+Rails.configuration.yeti_web['versioning_disable_for_models']&.each do |model_name|
+  PaperTrail.request.disable_model(model_name.constantize)
```
Fivell, line 6: @Skumring  @gigorok  from paper trail  docs, 

This setting, as with all PaperTrail.request settings, affects only the current request, not all threads.

Fivell: maybe somehow it works, but at least confusing for me , in my opinion it expected to work like
has_paper_trail if -> { paper_trail_enabled_for_model? }. 


```
def paper_trail_enabled_for_model?
   Rails.configuration.yeti_web['versioning_disable_for_models'].include?(self.model_name.to_s)
end 
```

gigorok: following specs it works as we expected. PaperTrail.request.disable_model allows us incapsulate logic in one place without modification models. maybe you can suggest some specs to check that ?

Fivell: @gigorok Im just worried about situation where RequestStore can be cleaned up in request store middleware and after request completed initializer will loose it logic

Fivell: or in other 3d party gem

Fivell: @gigorok  seems the most safe way is not to call has_paper_trail for model in general if config includes model name and disable all the versioning ui for such models

Fivell: @gigorok  seems the most safe way is not to call has_paper_trail for model in general if config includes model name and disable all the versioning ui for such models

gigorok: @Fivell agree, your more explicit solution will be better

app/models/billing/contact.rb
```diff
@@ -20,7 +20,7 @@
 
 class Billing::Contact < Yeti::ActiveRecord
   self.table_name = 'notifications.contacts'
-  has_paper_trail class_name: 'AuditLogItem'
+  has_paper_trail class_name: 'AuditLogItem' if Rails.configuration.yeti_web['versioning_disable_for_models'].exclude?(name)
```
gigorok, line 23: could you incapsulate 
`has_paper_trail class_name: 'AuditLogItem' if Rails.configuration.yeti_web['versioning_disable_for_models'].exclude?(name)` in some module ?

Fivell (approved):
@Skumring looks good, thanks

gigorok:
@Skumring awesome, thanks

Issue #846, opened 2021-01-25
fix floating failed tests


senid231:
```
1) BatchUpdateForm::Dialpeer check validation when all fields filled with valid values should pass validation
     Failure/Error: option_name = page.find(:xpath, ".//option[@value = '#{value}']").text
     
     Capybara::ElementNotFound:
       Unable to find xpath ".//option[@value = '1']" within #<Capybara::Node::Element tag="select" path="//HTML[1]/BODY[1]/DIV[8]/FORM[1]/DIV[1]/UL[1]/LI[6]/SELECT[1]">
     # ./spec/support/helpers/feature_test_helper.rb:8:in `block in select_by_value'
     # ./spec/support/helpers/feature_test_helper.rb:7:in `select_by_value'
     # ./spec/features/routing/dialpeers/batch_update_spec.rb:92:in `block (2 levels) in <top (required)>'
     # ./spec/features/routing/dialpeers/batch_update_spec.rb:227:in `block (2 levels) in <top (required)>'
     # ./spec/features/routing/dialpeers/batch_update_spec.rb:244:in `block (5 levels) in <top (required)>'
     # ./spec/features/routing/dialpeers/batch_update_spec.rb:243:in `block (4 levels) in <top (required)>'

rspec ./spec/features/routing/dialpeers/batch_update_spec.rb:242
```

Skumring:
```
Failures:

  1) Copy Rateplan action success when "Send quality alarms to" has values creates new Rateplan with identical fields, except UUID
     Failure/Error:
       create(:rateplan,
              send_quality_alarms_to: send_to_ids).reload # after_save

     ActiveRecord::RecordInvalid:
       Validation failed: Send quality alarms to is invalid
     Shared Example Group: :cloned_rateplan_is_valid called from ./spec/features/routing/rateplans/copy_rateplan_spec.rb:56
     # ./spec/features/routing/rateplans/copy_rateplan_spec.rb:26:in `block (3 levels) in <top (required)>'

  2) Dialpeers PUT /api/rest/admin/dialpeers/:id update values
     Failure/Error: contact.update!(email: email) if @email

     ActiveRecord::NotNullViolation:
       PG::NotNullViolation: ERROR:  null value in column "item_type" of relation "versions" violates not-null constraint
       DETAIL:  Failing row contains (287, null, null, null, null, null, null, null, null, 79565).
       : INSERT INTO "versions" DEFAULT VALUES RETURNING "id"
[...]

PR #854, opened 2021-01-28
Replace Coverall on SimpleCov and cofigure it

Skumring:
Fixes https://github.com/yeti-switch/yeti-web/issues/846

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,7 +35,7 @@ jobs:
             .bundle
             pgq-processors/.bundle
           key: gems-${{runner.os}}-${{hashFiles('Gemfile.lock')}}-${{hashFiles('pgq-processors/Gemfile.lock')}}
- 
+
   lint:
     name: Rubocop and bundle audit
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -125,13 +125,13 @@ jobs:
...
-#          github-token: ${{secrets.github_token}}
-#          flag-name: run-${{matrix.ci_node_index}}
-#          parallel: true
+      - name: SimpleCov Parallel
+        uses: coverallsapp/github-action@master
+        with:
+          path-to-lcov: "./coverage/lcov.info"
+          github-token: ${{secrets.github_token}}
+          flag-name: run-${{matrix.ci_node_index}}
+          parallel: true
```
dmitry-sinina, line 135: **finalize-coveralls** also should be uncommented, because we are using parrallelization

Skumring: @dmitry-sinina ok, I will do this tomorrow.

Skumring:
@dmitry-sinina, @gigorok looks like SimpleCov work correctly and statistics uploading to https://coveralls.io/builds/36789927 and when we merge changes to the master the coverall badge in Readme should be updated. But why failed `Coveralls - run-0` and `Coveralls - run-3` I don't understand, because I don't see any errors.

PR #868, opened 2021-02-06
Adding attrs to support registration gateway creation

thiadmer:
In addition to PR #829 I'd like to add a few other attributes to gateway and customerauth admin api to support creation of registration trunks.


dmitry-sinina:
@thiadmer could you also fix tests to use these attributes?

thiadmer:
Gateway attrs already covered by jsonapi_extract_fields_data, some others added, not sure why last rspec tests failing though

PR #881, opened 2021-02-15
fix API forgery origin check

gigorok:
API specific controllers should skip unnecessary layers (forgery protection, HTML rendering, translations, etc). so let's inherit them from ActionController::API instead of full featured ActionController::Base

also we need to upgrade knock gem as latest version already has solution on top of ActionController::API

### DIFF
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class ApiController < ActionController::Base
+class ApiController < ActionController::API
   around_action :db_logging
 
   def db_logging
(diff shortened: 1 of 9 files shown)


app/controllers/api_controller.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class ApiController < ActionController::Base
+class ApiController < ActionController::API
   around_action :db_logging
```
Fivell, line 4: need to update knock

gigorok: 👍

Fivell (approved):

Issue #882, opened 2021-02-16
add country, network and network prefix to cdr exports


Fivell:
need to create countries, networks and network_prefixes in cdr database
system should periodically synchronize (once a day) data between databases using new cron job(copy from routing database to CDR db)
(network prefixes can be big table so maybe it is better to read and insert in batches)

so job should just transactionally truncate tables and insert actual data,
cdr export should have ability to support new columns ( country name, network name, network prefix, maybe their ids as well, not sure)

PR #902, opened 2021-02-23
cdr export improve

Ivanov-Anton:
solve #882

Every day at 2 a.m. the tables will be synchronized by job named `Jobs::SyncDatabaseTables`

This was done in order to display the country name in CSV


Fivell:
@Ivanov-Anton  👍 , @dmitry-sinina  2 things here to discuss
1) schema where to put this new tables in cdr database (sys vs cdr vs something else) 
2) readonly admin ui to render countries and networks from cdr db since currently it is invisible for admin

dmitry-sinina:
> * schema where to put this new tables in cdr database (sys vs cdr vs something else)

It is better to create dedicated scheme **external_data**

>readonly admin ui to render countries and networks and prefixes from cdr db since currently it is invisible for admin

I think we don't need additional UI for same data located in othed DB. This feature completely internal stuff to solve internal problems(joining data from other database) so addition UI will confuse customers.

Issue #889, opened 2021-02-18
ActionView::Template::Error app/admin/reports/customer_traffic_data_by_vendors.rb in block (3 levels) in <main>


dmitry-sinina:
```
NoMethodError: undefined method `delete' for :red:Symbol
  from active_admin/views/components/status_tag.rb:32:in `build'
  from arbre/element/builder_methods.rb:30:in `block in build_tag'
  from arbre/context.rb:92:in `with_current_arbre_element'
  from arbre/element/builder_methods.rb:49:in `with_current_arbre_element'
  from arbre/element/builder_methods.rb:26:in `build_tag'
  from arbre/element/builder_methods.rb:39:in `insert_tag'
  from arbre/element/builder_methods.rb:14:in `status_tag'
  from app/admin/reports/customer_traffic_data_by_vendors.rb:60:in `block (3 levels) in <main>'
  from lib/active_admin/views/components/table_for.rb:140:in `render_data'
  from lib/active_admin/views/components/table_for.rb:134:in `block in build_table_cell'
  from arbre/element/builder_methods.rb:31:in `block in build_tag'
  from arbre/context.rb:92:in `with_current_arbre_element'
  from arbre/element/builder_methods.rb:49:in `with_current_arbre_element'
  from arbre/element/builder_methods.rb:26:in `build_tag'
  from arbre/element/builder_methods.rb:39:in `insert_tag'
  from arbre/element/builder_methods.rb:14:in `td'
  from lib/active_admin/views/components/table_for.rb:133:in `build_table_cell'
  from lib/active_admin/views/components/table_for.rb:59:in `block (2 levels) in column'
  from arbre/context.rb:92:in `with_current_arbre_element'
  from arbre/element/builder_methods.rb:49:in `with_current_arbre_element'
  from lib/active_admin/views/components/table_for.rb:58:in `block in column'
  from draper/delegation.rb:10:in `each'
  from draper/delegation.rb:10:in `each_with_index'
  from draper/delegation.rb:10:in `each_with_index'
  from lib/active_admin/views/components/table_for.rb:57:in `column'
  from arbre/element.rb:177:in `method_missing'
  from lib/active_admin/views/index_as_table.rb:38:in `column'
  from app/admin/reports/customer_traffic_data_by_vendors.rb:52:in `block (2 levels) in <main>'
  from lib/active_admin/views/index_as_table.rb:62:in `instance_exec'
[...]

PR #911, opened 2021-03-01
status_tag method was called incorrectly

Ivanov-Anton:
solve #889

### DIFF
--- a/app/admin/reports/customer_traffic_data_by_vendors.rb
+++ b/app/admin/reports/customer_traffic_data_by_vendors.rb
@@ -57,7 +57,7 @@ ActiveAdmin.register Report::CustomerTrafficDataByVendor, as: 'CustomerTrafficDa
       if row.vendor
         auto_link(row.vendor)
       else
-        status_tag('Rejected', :red)
+        status_tag('Rejected', class: :red)
       end
       # row.vendor.try(:display_name) || 'Rejected calls'
     end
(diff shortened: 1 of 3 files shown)


lib/active_admin/views/attributes_table.rb
```diff
@@ -4,7 +4,7 @@ module ActiveAdmin
   module Views
     class AttributesTable
       def bool_row(attribute, *args)
-        row(attribute, *args) { |model| model.send(attribute) ? status_tag('yes', :ok) : status_tag('no') }
+        row(attribute, *args) { |model| model.send(attribute) ? status_tag('yes', class: :ok) : status_tag('no') }
```
Fivell, line 7: @Ivanov-Anton  seems bool_row is not used anymore, better to delete this method

Ivanov-Anton: Seems whole class can be removed


[link](https://github.com/activeadmin/activeadmin/commit/3efb359533df46e76bb4355a02ad31e727fb0da3#diff-0700093aaea7d2d82f67221074c1e94fae30a568354fa56f3d9e46cf9c04e60c)

Ivanov-Anton: Removed in separate [PR](https://github.com/yeti-switch/yeti-web/pull/912)

Issue #935, opened 2021-04-03
It is not possible disable Invoice automatic generation


dmitry-sinina:
It is not possible to disable invoice autogeneration:

![image](https://user-images.githubusercontent.com/12762053/113483663-46064080-94ad-11eb-833a-febb4a081f4b.png)

Expected behaviour: 

admin should be  able to select empty option
no additional changes required, 
test: capybara test to switch from non none to none

PR #940, opened 2021-04-13
 add none option to select

BigG1947:
refs #935


Fivell:
@BigG1947  try to use allow single deselect chosen feature, issue was updated

dmitry-sinina:
why not `include_blank: Disable`?

PR #950, opened 2021-04-26
add routing_tag_ids chosen select to batch form

### LINKED ISSUE #375: Ability to update routing tags for Dialpeers and Destination using batch update form
Routing->Dialpeers->Collection Actions->Update batch

image

refs #738 

chosen multi-select should be used

BigG1947:
closes #375

### DIFF
--- a/app/assets/javascripts/build_tags.js
+++ b/app/assets/javascripts/build_tags.js
@@ -1,13 +1,19 @@
 // Build chosen field for routing tag ids select on update batch action form
 
 $(document).on('mass_update_modal_dialog:after_open', function (event, form) {
-    var tagsChosen = $(form).find("select[name='routing_tag_ids']").attr({'name': 'routing_tag_ids[]', 'class': 'chosen', 'id': 'routing_tag_ids'});
+    var tagsChosen = $(form).find("select[name='routing_tag_ids']").prop({
+        'name': 'routing_tag_ids[]', 
+        'class': 'chosen', 
+        'id': 'routing_tag_ids',
+        'value': '', // reset default value
+        'multiple': true
+    });
 
     if (tagsChosen.length === 0) {
         return;
     }
 
-    var hidden = $('<input>').attr({
+    var hidden = $('<input>').prop({
         'type': 'hidden',
         'id': 'routing_tag_ids',
         'name': 'routing_tag_ids',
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/build_tags.js
```diff
@@ -0,0 +1,46 @@
...
+    
+    var tagChosen = $('select#routing_tag_ids.chosen');
+    var disabled = tagChosen.prop('disabled')
+
+    if (disabled) { 
+        $("[name='routing_tag_ids']").prop('disabled', true); 
+    }else if(!disabled && tagChosen.val() == ''){
+        $("[name='routing_tag_ids']").prop('disabled', false); 
+    }
+});
```
senid231, line 46: 1. why did you add this handler on document instead of add it in upper function?
2. it looks like uglier duplicate of above code inside `tagsChosen.change({hidden: hidden}, function(event){`

senid231 (changes requested):

app/assets/javascripts/build_tags.js
```diff
@@ -0,0 +1,46 @@
...
+
+    tagsChosen.chosen({no_results_text: "No results matched", width: '240px', search_contains: true, allow_single_deselect: true});
+
+    tagsChosen.change({hidden: hidden}, function(event){
+        if ($(event.target).val() != ''){
+            event.data.hidden.prop('disabled', true);
+        }else{
+            event.data.hidden.prop('disabled', false);
+        }
+    });
```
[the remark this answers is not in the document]
BigG1947: > not when tagsChosen changed

Also I need disable hidden field when user select some tags

app/assets/javascripts/build_tags.js
BigG1947: > 1. why did you add this handler on document instead of add it in upper function?
> 2. it looks like uglier duplicate of above code inside `tagsChosen.change({hidden: hidden}, function(event){`

This is not same event. This event triggered when user changed status of anyone checkbox.
When checkbox near tagChosen is not checked - hidden field must be disabled.
When checkbox is checked and tagChosen is not have any tags - hidden field must be enabled.

Fivell (approved):

spec/features/routing/destinations/batch_update_spec.rb
```diff
@@ -163,6 +165,12 @@
       check :Short_calls_limit
       fill_in :short_calls_limit, with: assign_params[:short_calls_limit]
     end
+
+    if assign_params.key? :routing_tag_ids
+      check :Routing_tag_ids
+      routing_tags.select { |tag| assign_params[:routing_tag_ids].include? tag.id.to_s }
+                  .each { |tag| fill_in_chosen 'routing_tag_ids[]', with: tag.name, multiple: true, visible: false }
```
senid231, line 172:
```
page.scroll_to find_field('routing_tag_ids[]')
assign_params[:routing_tag_ids].each do |tag_id|
  fill_in_chosen 'routing_tag_ids[]', with RoutingTag.find(tag_id).name, multiple: true
end
```

it will be more readable - and no need for `visible: false` here

spec/features/routing/destinations/batch_update_spec.rb
```diff
@@ -183,5 +191,16 @@
         end.to have_enqueued_job(AsyncBatchUpdateJob).on_queue('batch_actions').with 'Routing::Destination', be_present, assign_params, be_present
       end
     end
+
+    context 'when routing tag ids field is empty' do
+      let(:assign_params) { { routing_tag_ids: '' } }
```
senid231, line 196: value should be empty array

`let(:assign_params) { { routing_tag_ids: [] } }`

senid231 (changes requested):

PR #982, opened 2021-05-24
add new filters for cdr and cdr export resources

### LINKED ISSUE #980: cdr resource and cdr export resource new filters required
### external systems can use they own countries identifications, only ISO can be used in such cases
such filter should be supported for API **exports** and **CDR rest api** 
`dst_country_iso_equals=US` and `src_country_iso_equals=US`
before filtering in database it should  fetch id by iso code in countries table and return with 400  `InvalidFilterValue` if no such iso found

for export InvalidFieldValue exception for fields  attribute if not valid iso

### external system can have ability to filter cdrs with specific routing_tag_id for  **exports**, **rest api** , **admin ui**
`routing_tag_ids_includes=4` `routing_tag_ids_exludes=5`

`routing_tag_ids_includes` should convert to `WHERE 4= ANY(routing_tag_ids)` 

`routing_tag_ids_excludes` should convert to `WHERE NOT(5 = ANY(routing_tag_ids))` 

routing_tag_ids_blank (true/false)  should be converted to 

`WHERE routing_tag_ids IS NULL OR routing_tag_ids = '{}' ` 

or 
`WHERE NOT (routing_tag_ids IS NULL OR routing_tag_ids = '{}' )`

BigG1947:
closes #980


dmitry-sinina:
could you also add acceptance test to generate API documentation - currently there is no API documentation for cdr-exports resource

dmitry-sinina:
https://demo.yeti-switch.org/doc/api/admin/index.html

BigG1947:
> could you also add acceptance test to generate API documentation - currently there is no API documentation for cdr-exports resource
> https://demo.yeti-switch.org/doc/api/admin/index.html

I did something similar, but this new post creation page does not provide information on what filters are required for cdr_export (for example, time_start_gteq and time_start_lteq). When I try to point this out, I have a problem with how the filter should be named in the request body. Usually all parameters have the following pattern: "parameter-as-is-name", but for 'cdr export filters' parameters must be named with underscore, for example: "for_cdr_filter_parameter_name".

Screenshot_2021-05-28 Admin API
Screenshot_2021-05-28 Cdr Exports API
Screenshot_2021-05-28 Cdr Exports API(1)

app/resources/api/rest/admin/cdr/cdr_export_resource.rb
```diff
@@ -15,6 +15,26 @@ def filters
     _model.filters.as_json
   end
 
+  def filters=(f)
+    src_country_iso = f.delete('src_country_iso_eq')
+    src_country = System::Country.find_by(iso2: src_country_iso)
```
Fivell, line 20: @BigG1947 use logic conditionally
```
if f.has_key?('src_country_iso_eq')
   ....
end

if f.has_key?('dst_country_iso_eq')
   ....
end
```

PR #1081, opened 2022-01-04
Add RX SSRC and TX SSRC filters to Rtp Statistics page

### LINKED ISSUE #1032: Feature Request: Add filters RX SSRC and TX SSRC to "Rtp Statistics" Admin Page
Add new filters to "Rtp Statistics" admin page:

RX SSRC
TX SSRC

Filtering options:
Equals

Ivanov-Anton:
closes #1032


dmitry-sinina:
this filter should accept **hex** values as well as integer
for example `0x123345` should be converted to `74565` and then filter should be applied

dmitry-sinina:
also **hex** should be default selection

PR #1102, opened 2022-02-12
add chosen-ajax-fillable, DRY association ajax inputs/filters

senid231:
* add chosen-ajax-fillable for account inputs
* DRY association ajax filter/input


Fivell (approved):
👍

PR #1107, opened 2022-02-19
admin UI active nodes - do not send sentry error when node is down

senid231:


senid231:
https://github.com/yeti-switch/yeti-web/pull/1108 will fix floating test

PR #1145, opened 2022-06-22
rateplans and routing plans copy fixes

Fivell:


senid231:
`spec/support/helpers/chosen.rb`

```ruby
def have_selected_chosen(label, multiple: false, value:)
      container_selector = chosen_container_selector(label)
      if multiple
        selector = "#{container_selector} .chosen-choices > .search-choice > span"
        satisfy do |page|
          expect(page).to have_selector(selector, count: value.size)
          value.each do |val|
            expect(page).to have_selector(selector, exact_text: val.to_s)
          end
        end
      else
        selector = "#{container_selector} .chosen-single > .span"
        have_selector(selector, exact_text: value.to_s)
      end
    end
    ```

usage
```ruby
expect(page).to have_selector_chosen 'Rate groups', multiple: true, value: ['one', 'two']
expect(page).to have_selector_chosen 'Rate group', value: 'one'
```

senid231:
Routing plans
image
image


Rate plans
image
image

PR #1207, opened 2022-11-16
fix customer api charts endponts to return timezone

dmitry-sinina:


senid231:
vulnerabilities fixed in separate PR  
https://github.com/yeti-switch/yeti-web/pull/1225

PR #1240, opened 2023-02-26
improve tests for billing.bill_cdr_batch

senid231:


spec/sql/billing/bill_cdr_batch_spec.rb
```diff
@@ -2,35 +2,17 @@
 
 RSpec.describe 'billing.bill_cdr_batch' do
   subject do
-    SqlCaller::Yeti.select_all('SELECT * FROM billing.bill_cdr_batch(?,?)', batch_id, batch_data).map(&:deep_symbolize_keys)
+    SqlCaller::Yeti.execute('SELECT * FROM billing.bill_cdr_batch(?,?)', batch_id, batch_data.to_json)
```
senid231, line 5: use `execute` instead of `select_all` because function has no return value

spec/sql/billing/bill_cdr_batch_spec.rb
```diff
@@ -88,30 +70,58 @@
         term_gw_id: gw.id,
         duration: -11 # should be skipped in stats
       }
-    ].to_json.to_s
```
senid231, line 91: `batch_data` being conerted to json in the `subject`.  
FYI no need to call `to_json.to_s`, the `to_json` method already return a string

spec/sql/billing/bill_cdr_batch_spec.rb
```diff
@@ -88,30 +70,58 @@
...
+  it 'updates customer accounts' do
+    customer_acc1_balance = customer_acc1.balance
+    # destination_reverse_billing=false and customer_price > 0
+    customer_acc1_spending = batch_data[0][:customer_price] + batch_data[1][:customer_price]
+    # destination_reverse_billing=true and customer_price > 0
+    customer_acc1_income = batch_data[2][:customer_price]
+
+    customer_acc2_balance = customer_acc2.balance
+    # destination_reverse_billing=false and customer_price > 0
+    customer_acc2_spending = batch_data[4][:customer_price]
```
senid231, line 85: calculate expected balance before calling `subject` with explanation.  
it's much more useful for other developers, than some magic numbers.  
such tests are easy to copy+paste into new contexts and change according to different test cases

spec/sql/billing/bill_cdr_batch_spec.rb
```diff
@@ -88,30 +70,58 @@
...
+    # dialpeer_reverse_billing=true and vendor_price > 0
+    vendor_acc2_spending = batch_data[3][:vendor_price]
     subject
-    expect(Account.find(customer_acc1.id).balance).to eq -6.001 - 10
-    expect(Account.find(customer_acc2.id).balance).to eq -11
-    expect(Account.find(vendor_acc1.id).balance).to eq 1100 + 99
-    expect(Account.find(vendor_acc2.id).balance).to eq 1120 - 0.99999
+    expect(vendor_acc1.reload.balance).to eq(vendor_acc1_balance + vendor_acc1_income)
+    expect(vendor_acc2.reload.balance).to eq(vendor_acc2_balance + vendor_acc2_income - vendor_acc2_spending)
+  end
```
senid231, line 105: single test split into 3 for readability

spec/sql/billing/bill_cdr_batch_spec.rb
```diff
@@ -88,30 +70,58 @@
...
+  it 'updates statistic' do
+    subject
+    expect(dp.reload.statistic).to have_attributes(
+                                     calls: 3,
+                                     calls_success: 2,
+                                     calls_fail: 1,
+                                     total_duration: 32,
+                                     asr: 0.6666667,
+                                     acd: 16
+                                   )
```
senid231, line 116: 1. instead of `Dialpeer.find(dp.id)` you can just call `dp.reload` that will refetch attributes/associations from database.
2. better use `have_attributes` matcher to check several attributes (getter methods) of single object, because it's easy to read and extend.

PR #1243, opened 2023-03-02
allow delete routing tags linked to applied pricelist items

BigG1947:


app/models/routing/routing_tag.rb
```diff
@@ -32,8 +32,13 @@ class Routing::RoutingTag < ApplicationRecord
...
-           ->(tag) { unscope(:where).where("? = ANY(#{table_name}.routing_tag_ids)", tag.id) },
+  has_many :active_rate_management_pricelist_items,
+           ->(tag) { unscope(:where).not_applied.where("? = ANY(#{table_name}.routing_tag_ids)", tag.id) },
+           class_name: 'RateManagement::PricelistItem',
+           autosave: false
+
+  has_many :applied_rate_management_pricelist_items,
+           ->(tag) { unscope(:where).applied.where("? = ANY(#{table_name}.routing_tag_ids)", tag.id) },
            class_name: 'RateManagement::PricelistItem',
            autosave: false
```
senid231, line 43: `applied_rate_management_pricelist_items` is not used anywhere so can be removed

app/models/routing/routing_tag.rb
```diff
@@ -69,7 +74,7 @@ def prevent_destroy_if_have_assosiations
       errors.add(:base, "Can't be deleted because linked to Rate Management Project(s) ##{project_ids.join(', #')}")
     end
 
-    pricelist_ids = rate_management_pricelist_items.pluck(Arel.sql('DISTINCT(pricelist_id)'))
+    pricelist_ids = active_rate_management_pricelist_items.pluck(Arel.sql('DISTINCT(pricelist_id)'))
```
senid231, line 77: here it will be nice to leave a comment on why we validate only active pricelist items  
```ruby
# we allow keeping deleted routing tag IDs in applied rate management pricelist items
# because it does not break anything and such records are used only as a temporary history
```

senid231 (changes requested):

senid231 (approved):

PR #1292, opened 2023-06-07
Add customer auth stats last24h to prometheus

senid231:


senid231:
closed if favor of https://github.com/yeti-switch/yeti-web/pull/1295

PR #1315, opened 2023-07-10
fix error when destroying numbelist assigned to customer auth

BigG1947:


dmitry-sinina:
@BigG1947 Numberlists also may be linked to gateways

PR #1334, opened 2023-08-31
#1333, store secret key value inside secrets.yml

Ivanov-Anton:
instead of storing secret_key_base in the initializer the rails (from v4.1) suggests to use config/secrets.yml

https://guides.rubyonrails.org/4_1_release_notes.html#config-secrets-yml


gigorok (approved):

Ivanov-Anton:
@dmitry-sinina Thank you so much, we can create the new tag

PR #1345, opened 2023-09-27
Add new filters to the CDR Exports

Skumring:
Add additional CDR export filters: customer_auth_external_type_eq, customer_auth_external_type_not_eq, customer_auth_external_id_in (supports array), src_country_iso_in (supports array), dst_country_iso_in (supports array)

filters
cdr_export_show

### DIFF
--- a/app/admin/cdr/cdr_exports.rb
+++ b/app/admin/cdr/cdr_exports.rb
@@ -108,7 +108,53 @@ ActiveAdmin.register CdrExport, as: 'CDR Export' do
   end
 
   permit_params :callback_url,
-                filters: CdrExport::FiltersModel.attribute_types.keys.map(&:to_sym),
+                filters: [
+                  :time_start_gteq,
+                  :time_start_lteq,
+                  :time_start_lt,
+                  :customer_id_eq,
+                  :customer_external_id_eq,
+                  :customer_acc_id_eq,
+                  :customer_acc_external_id_eq,
+                  :vendor_id_eq,
+                  :vendor_external_id_eq,
+                  :vendor_acc_id_eq,
+                  :vendor_acc_external_id_eq,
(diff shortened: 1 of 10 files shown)


dmitry-sinina:
Is is possible to add "Copy feature". To create new CDR export based on values from previous one?

dmitry-sinina:
There is copy button in every object show page, for example customer auth:

image

But by some reasons there is no such button in CDR export. It would be quite useful to have it to create same CDR export where only date range is different.

Skumring:
> There is copy button in every object show page, for example customer auth:
> 
> image
> But by some reasons there is no such button in CDR export. It would be quite useful to have it to create same CDR export where only date range is different.

Date range - means created_at&updated_at timestamps of CdrExport or you mean filters: `time_start_gteq` & `time_start_lteq` & `time_start_lt` inside json attribute `filters` in CdrExport

gigorok:
> Is is possible to add "Copy feature". To create new CDR export based on values from previous one? Same as we have for other objects.

@dmitry-sinina could you create a new issue? "Copy feature" will be implemented in another scope of improvements

app/admin/cdr/cdr_exports.rb
```diff
@@ -225,6 +271,26 @@ def build_new_resource
                input_html: { class: 'chosen' },
                required: false
       ff.input :term_gw_external_id_eq, required: false
+      ff.input :customer_auth_external_type_eq, required: false
+      ff.input :customer_auth_external_type_not_eq, required: false
+      ff.input :customer_auth_external_id_in,
+               as: :select,
+               multiple: true,
+               collection: CustomersAuth.order(:name).pluck(:name, :external_id),
```
gigorok, line 279: could it be implemented using Ajax? possible big amount of customers auth records

PR #1347, opened 2023-09-28
#1200, Add request IP and meta to API log model

### LINKED ISSUE #1200: Add request IP and meta to API log model
Additionally, add the possibility to write meta info to the API log

store additional info about the current user such as (current_customer_id, remote_ip)

## Technical info:

1. add a `meta` column for `logs.api_requests` table nullable, default: {}, with `jsonb` type
2. fill in this new column when "process_action.action_controller" event occurs with current_customer_id and remote_ip info
3. render this info in Admin UI

Ivanov-Anton:
add meta column with `jsonb` type to logs.api_requests table
add remote_ip column with `string` type to logs.api_requests table

Added this filter on the API Log index page:
- "Remote IP" string by remote_ip column

#### store remote IP when API/meta info is used (any endpoint that starts with '/api/')

closed #1200

### DIFF
--- a/app/admin/logs/api_logs.rb
+++ b/app/admin/logs/api_logs.rb
@@ -16,7 +16,7 @@ ActiveAdmin.register Log::ApiLog, as: 'ApiLog' do
 
   controller do
     def scoped_collection
-      super.select('id, created_at, status, method, path, db_duration, page_duration')
+      super.select('id, created_at, status, method, path, db_duration, page_duration, remote_ip')
     end
 
     def find_resource
@@ -37,6 +37,7 @@ ActiveAdmin.register Log::ApiLog, as: 'ApiLog' do
   filter :response_body
   filter :request_headers
   filter :response_headers
+  filter :remote_ip_eq_inet, as: :string, label: 'Remote IP'
 
   index do
     id_column
@@ -46,6 +47,7 @@ ActiveAdmin.register Log::ApiLog, as: 'ApiLog' do
     column :path
     column :db_duration
     column :page_duration
+    column :remote_ip
   end
 
   show do
@@ -84,6 +86,12 @@ ActiveAdmin.register Log::ApiLog, as: 'ApiLog' do
           l.response_body
         end
       end
+      row :meta do |l|
+        pre do
+          l.meta
+        end
+      end
+      row :remote_ip
     end
     active_admin_comments
   end
--- a/app/models/log/api_log.rb
+++ b/app/models/log/api_log.rb
@@ -8,10 +8,12 @@
 #  action           :string
 #  controller       :string
 #  db_duration      :float
+#  meta             :jsonb
 #  method           :string
 #  page_duration    :float
 #  params           :text
 #  path             :string
+#  remote_ip        :inet
 #  request_body     :text
 #  request_headers  :text
 #  response_body    :text
@@ -36,8 +38,13 @@ class Log::ApiLog < ApplicationRecord
   self.pg_partition_depth_future = 3
 
   scope :failed, -> { where('status >= ?', 400) }
+  scope :remote_ip_eq_inet, ->(value) { where('remote_ip >>= ?', value) }
 
   def display_name
     id.to_s
+  end
+
+  def self.ransackable_scopes(_auth_object = nil)
+    %i[remote_ip_eq_inet]
   end
 end
--- a/config/initializers/instrumentation_notification.rb
+++ b/config/initializers/instrumentation_notification.rb
@@ -12,6 +12,8 @@ module ActionController
     end
 
     def process_action(*args)
+      return super unless request.path.start_with?('/api/')
+
       raw_payload = {
         controller: self.class.name,
         action: action_name,
--- a/config/initializers/instrumentation_notification.rb
+++ b/config/initializers/instrumentation_notification.rb
@@ -15,6 +15,7 @@ module ActionController
       raw_payload = {
         controller: self.class.name,
         action: action_name,
+        meta: meta,
         params: request.filtered_parameters,
         format: request.format.try(:ref),
         method: request.method,
@@ -43,6 +44,9 @@ ActiveSupport::Notifications.subscribe 'process_action.action_controller' do |_n
 
     Log::ApiLog.create do |api_request|
       debug_mode = payload[:debug_mode]
+
+      api_request.meta = payload[:meta]
+      api_request.remote_ip = payload[:remote_ip]
 
       if payload[:status].nil? && payload[:exception].present?
         payload[:status] = ActionDispatch::ExceptionWrapper.status_code_for_exception(payload[:exception].class)
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -25446,7 +25446,9 @@ CREATE TABLE logs.api_requests (
     request_body text,
     response_body text,
     request_headers text,
-    response_headers text
+    response_headers text,
+    meta jsonb,
+    remote_ip inet
 )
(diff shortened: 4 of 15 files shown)


db/structure.sql
```diff
@@ -25446,7 +25446,9 @@ CREATE TABLE logs.api_requests (
     request_body text,
     response_body text,
     request_headers text,
-    response_headers text
+    response_headers text,
+    meta jsonb DEFAULT '{}'::jsonb,
```
dmitry-sinina, line 25450: can it be null by default?

Ivanov-Anton: OK

db/structure.sql
```diff
@@ -25446,7 +25446,9 @@ CREATE TABLE logs.api_requests (
...
+    remote_ip character varying
```
dmitry-sinina, line 25451: also *inet* for remote_ip will save some space.

Ivanov-Anton:
@dmitry-sinina do you know how to create an SQL request that will be searched within meta attributes something like this `where('meta: { current_customer_id: 2 }')`?

I would like to add a filter in the index page of the API Log page to have the ability to search logs within specific customer_id.

dmitry-sinina:
```
SELECT jsonb_extract_path_text('{"current_customer_id": 2 }'::jsonb, 'current_customer_id')='2';
 ?column? 
----------
 t
(1 row)
```

dmitry-sinina:
Other option:
```
SELECT {"current_customer_id": 2,"test":222 }'::jsonb @> '{"current_customer_id": 2,"test":222 }'::jsonb
```

dmitry-sinina:
@Ivanov-Anton looks like new postgresql causing have different behavior, and same fix for tests should be applied as we done before:
https://github.com/yeti-switch/yeti-web/pull/1340/files#diff-e05d722809b51b6d20f014e0130d0044a9fb18c7b69d2a9804416e674f0ef125R28

to ignore order or records.

Ivanov-Anton:
@dmitry-sinina OK I want to add one more filter on the API Log index page to have ability to filter logs by Contractor (ajax filter)

dmitry-sinina:
also for IP better to use filter like this:
```
 SELECT '192.168.0.0/16'::inet >>= 192.168.1.5/32::inet
;
 ?column? 
----------
 t
(1 row)

```
so entering network with mask will allow match any IP from this network, and without mask it will work as equal:
```
 # SELECT '192.168.1.5'::inet >>= '192.168.1.5/32'::inet
;
 ?column? 
----------
 t
(1 row)

Time: 0,835 ms
```

dmitry-sinina:
@Ivanov-Anton still draft?

app/admin/logs/api_logs.rb
```diff
@@ -37,6 +47,8 @@ def find_resource
   filter :response_body
   filter :request_headers
   filter :response_headers
+  filter :remote_ip_eq_inet, as: :string, label: 'Remote IP'
+  filter :current_customer_id_equal, as: :string, label: 'Current Customer ID Equal'
```
gigorok, line 51: no need to add filtering by meta, it can be anything info, not necessarily the customer. filtering by remote_ip will be enough

config/initializers/instrumentation_notification.rb
```diff
@@ -24,6 +27,7 @@ def process_action(*args)
                  'unknown'
                     end)
       }
+      raw_payload[:meta] = { current_customer_id: current_customer_id } if current_customer_id
```
gigorok, line 30: write `meta` from controller scope, not hardcode customer here

app/admin/logs/api_logs.rb
```diff
@@ -84,6 +85,12 @@ def find_resource
           l.response_body
         end
       end
+      row :meta do |l|
+        pre do
+          l.meta
+        end
+      end
+      row :remote_ip
```
gigorok, line 93: add remote_ip to index page

config/initializers/instrumentation_notification.rb
```diff
@@ -12,9 +12,12 @@ def append_info_to_raw_payload(raw_payload)
     end
 
     def process_action(*args)
+      meta = respond_to?(:meta) ? self.meta : nil
```
gigorok, line 15: could we not to check method definition?

app/admin/logs/api_logs.rb
```diff
@@ -76,14 +78,21 @@ def find_resource
       end
       row :response_headers do |l|
         pre do
-          l.response_headers
+          l.response_hea
+          ders
```
gigorok, line 82: remove extra whitespace

app/admin/logs/api_logs.rb
```diff
@@ -16,7 +16,7 @@
 
   controller do
     def scoped_collection
-      super.select('id, created_at, status, method, path, db_duration, page_duration')
+      super.select('id, created_at, status, method, path, db_duration, page_duration, remote_ip')
```
gigorok, line 19: do we need to select `meta`?

app/controllers/application_controller.rb
```diff
@@ -38,6 +38,10 @@ def info_for_paper_trail
     { ip: request.env['HTTP_X_REAL_IP'] || request.remote_ip }
   end
 
+  def meta
```
gigorok, line 41: no need to add meta to the application controller. only for API specific controllers

app/models/log/api_log.rb
```diff
@@ -36,8 +38,13 @@ class Log::ApiLog < ApplicationRecord
   self.pg_partition_depth_future = 3
 
   scope :failed, -> { where('status >= ?', 400) }
+  scope :remote_ip_eq_inet, ->(value) { where('remote_ip::inet >>= ?', value) }
```
gigorok, line 41: `remote_ip` is already `inet`. do we need type casting again?

spec/requests/api/instrumentation_notification_spec.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+RSpec.describe 'config/initializers/instrumentation_notification.rb' do
```
gigorok, line 3: remove this spec file. there is no information anymore about `current_customer_id`. better to add a spec to one of API request spec

app/admin/logs/api_logs.rb
Ivanov-Anton: no, we do not need to select the meta attribute in the index page, I think showing meta in the show page will be enough.

spec/requests/api/instrumentation_notification_spec.rb
Ivanov-Anton: @gigorok this changes is not safe and after apply this notice almost all test activeadmin feature test is failed. Maybe I suggest to add this line in process action(*args) method

```
return super unless request.path.start_with?('/api/')
```

The problem is next active admin generates controller and we have error that some_page_controller do not have meta method.

app/admin/logs/api_logs.rb
gigorok: why then selecting `remote_ip` is needed?

config/initializers/instrumentation_notification.rb
```diff
@@ -15,6 +15,8 @@ def process_action(*args)
       raw_payload = {
         controller: self.class.name,
         action: action_name,
+        remote_ip: request.remote_ip,
```
gigorok, line 18: looks like `remote_ip` is already present in the payload for API controllers. see app/controllers/concerns/with_payloads.rb. we can try to set `meta` here as well

PR #1357, opened 2023-10-20
#1276, Refactor AsyncBatchDestroyJob to increase efficiency

### LINKED ISSUE #1276: Bacth actions job validations for numberlists
image

image

Ivanov-Anton:
## Description

Update AsyncBatchDestroyJob to handle destroy failures

The AsyncBatchDestroyJob has been refactored to handle destroy failures, instead of raising errors. A warning is logged if a destroy operation fails, improving the robustness of the job. Corresponding changes have also been made to the related test cases.

## Additional links

closes #1276

### DIFF
--- a/app/jobs/async_batch_destroy_job.rb
+++ b/app/jobs/async_batch_destroy_job.rb
@@ -5,24 +5,40 @@ class AsyncBatchDestroyJob < ApplicationJob
   BATCH_SIZE = 1000
   queue_as 'batch_actions'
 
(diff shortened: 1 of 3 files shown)


dmitry-sinina:
@Ivanov-Anton I am not sure this is good idea. From my point of view we should not allow deletion of used numberlist.

dmitry-sinina:
I think we need some validation to fix #1276 
May be background job should skip such objects during deletion.

Issue #1346, opened 2023-09-28
Lack of Copy feature for CDR export


dmitry-sinina:
Usually for every object we have "Copy" button to create new object based on attributes of previous one., customer auth for example:

![image](https://github.com/yeti-switch/yeti-web/assets/12762053/7fbc9fd2-1a3b-47a8-89c1-e8d2aca11bc9)

But by some reasons there is no such feature on CDR export page, it would be very useful especially because there are lot of input filters in this page.

PR #1364, opened 2023-10-30
Copy feature for CDR export

AlexeyKasianenko:


app/admin/cdr/cdr_exports.rb
```diff
@@ -4,6 +4,20 @@
   menu parent: 'CDR', priority: 97
   actions :index, :show, :create, :new
 
+  acts_as_clone except: lambda { |params|
+    except = []
+    from = CdrExport.find(params[:from])
+    except << :customer_auth_external_id_in unless CustomersAuth.exists?(id: from.filters[:customer_auth_external_id_in])
```
dmitry-sinina, line 10: why we need this exception logic there?

senid231: removed

senid231:
done #1346

PR #1379, opened 2023-12-01
#1378, fill in "Customer auth external id in" field after form reloading

### LINKED ISSUE #1378: Bug inside "Cdr Export" form
Steps to reproduce:
1. Go to New CDR export page `CDR Export` form
2. Fill in customer_auth_external_id_in
3. Click on the "Create Cdr export" button

### Expected behavior:
1. There is a validation error
2. The previously set value is displayed in customer_auth_external_id_in

### Actual behavior:
1. There is a validation error
2. The previously set value is `NOT` displayed in customer_auth_external_id_in

Ivanov-Anton:
Bug fix form CDR -> CDR Export -> New

### Bug description:

When submitting the invalid form and after the form reloads and displays semantic errors the previously selected `customer_auth_external_id_in` field is not present which is incorrect.

  * detect previously selected value of `customer_auth_external_id_in`
  * AJAX search `CustomerAuth` with `external_id` NOT NULL `only`

### Additional info:

AJAX behavior was changed to add an extra where condition. This was done because the reported bug was still occurring when CustomerAuth didn't have an external_id. It's illogical to return Customer Auth records without `external_id`, as CDR Export filters data by this column see filter name (Customer auth external id in).

closes #1378


app/admin/cdr/cdr_exports.rb
```diff
@@ -282,11 +282,14 @@ def build_new_resource
                input_html: {
                  class: 'chosen-ajax',
                  data: {
-                   path: '/customers_auths/search_with_return_external_id?q[ordered_by]=name'
+                   path: '/customers_auths/search_with_return_external_id?q[ordered_by]=name&q[external_id_null]=false'
                  }
                },
-               collection: if params.dig(:q, :customer_auth_external_id_in)
-                             CustomersAuth.where(external_id: params.dig(:q, :customer_auth_external_id_in)).order(:name).pluck(:name, :external_id)
+               collection: if params.dig(:cdr_export, :filters, :customer_auth_external_id_in)
```
gigorok, line 288: instead of params checking try to get values like this `ff.object.customer_auth_external_id_in`

app/admin/cdr/cdr_exports.rb
```diff
@@ -282,11 +282,14 @@ def build_new_resource
                input_html: {
                  class: 'chosen-ajax',
                  data: {
-                   path: '/customers_auths/search_with_return_external_id?q[ordered_by]=name'
+                   path: '/customers_auths/search_with_return_external_id?q[ordered_by]=name&q[external_id_null]=false'
                  }
                },
-               collection: if params.dig(:q, :customer_auth_external_id_in)
-                             CustomersAuth.where(external_id: params.dig(:q, :customer_auth_external_id_in)).order(:name).pluck(:name, :external_id)
+               collection: if ff.object.customer_auth_external_id_in
```
gigorok, line 288: `customer_auth_external_id_in` is array? maybe we need to add `present?`?

PR #1399, opened 2023-12-27
YETIWEB-1398, Feature: Introduce API Log tags

### LINKED ISSUE #1398: api log tagging
Introduce API Log tags. So you can add tag/tags "api_log_tags" config to "yeti_web.yml" configuration file, and this value will be added to each API log record.

### Additional info:
After implementing this feature we need to create a new release 1.12.124

Ivanov-Anton:
Introduce API Log tags. So you can add tag/tags "api_log_tags" config to the "yeti_web.yml" configuration file, and this value will be added to each API log record.

Admin UI changes:
- added column tags for the index page of the API Log
- added column tags for show page of API Log
- added filter Tag Contains for the index page of API Log

closes #1398

## Screenshot:

Screenshot 2023-12-27 at 12 17 51

### DIFF
--- a/app/admin/logs/api_logs.rb
+++ b/app/admin/logs/api_logs.rb
@@ -16,7 +16,7 @@ ActiveAdmin.register Log::ApiLog, as: 'ApiLog' do
 
   controller do
     def scoped_collection
-      super.select('id, created_at, status, method, path, db_duration, page_duration, remote_ip')
+      super.select('id, created_at, status, method, path, db_duration, page_duration, remote_ip, tags')
     end
 
     def find_resource
@@ -27,6 +27,7 @@ ActiveAdmin.register Log::ApiLog, as: 'ApiLog' do
   filter :created_at, as: :date_time_range
   filter :path
   filter :method
+  filter :tag_eq, label: 'Tag Equals'
   filter :status
   filter :controller, as: :select, collection: proc { ApiControllers.list }, input_html: { class: :chosen }
   filter :action
@@ -44,6 +45,7 @@ ActiveAdmin.register Log::ApiLog, as: 'ApiLog' do
     column :created_at
     column :status
     column :method
+    column :tags, &:tags_separated_by_comma
     column :path
     column :db_duration
     column :page_duration
@@ -56,6 +58,7 @@ ActiveAdmin.register Log::ApiLog, as: 'ApiLog' do
       row :created_at
       row :status
       row :method
+      row :tags, &:tags_separated_by_comma
       row :path
       row :controller
       row :action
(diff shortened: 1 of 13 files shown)


app/admin/logs/api_logs.rb
```diff
@@ -27,6 +27,10 @@ def find_resource
   filter :created_at, as: :date_time_range
   filter :path
   filter :method
+  filter :tags_array_contains, as: :select,
+                               collection: proc { YetiConfig.api_log_tags&.uniq || [] },
```
dmitry-sinina, line 31: There you building list of all possible tags.

config/initializers/instrumentation_notification.rb
```diff
@@ -21,6 +21,7 @@ def process_action(*args)
         params: request.filtered_parameters,
         format: request.format.try(:ref),
         method: request.method,
+        tags: YetiConfig.api_log_tags.presence || [],
```
dmitry-sinina, line 24: There you have to add not tags defined in **api_log_tags**(which is all possible tags) but tags from other variable, like **api_log_add_tags**. In this way it will be possible to add different tags on  different application instances

Ivanov-Anton: so to implement this suggestion I want to know how to fill in this additional variable/method

here example of usage

```
def process_action(*args)
      return super unless request.path.start_with?('/api/')

      raw_payload = {
        controller: self.class.name,
        action: action_name,
        meta: meta,
        params: request.filtered_parameters,
        format: request.format.try(:ref),
        method: request.method,
        tags: api_log_add_tags,
        path: (begin
                      request.fullpath
               rescue StandardError
                 'unknown'
                    end)
      }
```
Would you like to define the api_log_add_tags method in these controllers?

```
          app/controllers/api/cryptomus_callbacks_controller.rb
          app/controllers/api/rest/admin/auth_controller.rb
          app/controllers/api/rest/customer/v1/auth_controller.rb
          app/controllers/api_controller.rb
```

or you want to do define a separate config section (`api_log_add_tags `) inside Yeti Web yml config file

dmitry-sinina: use case:

- Two instances of application
- Instance A, should add tagA to logs
- Instance B, should add tagB to logs
- When admin checking logs in UI he should be able to filter logs by tagA and tagB - it means UI should know all possible tags to build list in filter.

solution1:
- two config variables
- **api_log_tags** - to define all possible tags(will be used to build list in filters in UI). it will have same value on both instances( [tagA, tagB].
- **api_log_add_tags** - to define which tag will be added during writing API logs. it will have different values on each instance

solution2:
- one config variable
- no tag list in filter UI, just text input to allow enter any tag name manually. It is admin responsibility to know which value to enter
- one config variable to define which tag will be added during writing API logs. it will have different values on each instance

gigorok: @dmitry-sinina @Ivanov-Anton solution2 is more simple. I suggest to implement it

PR #1458, opened 2024-04-08
schedule rate changes for destinations

BigG1947:
image
image
image


app/admin/routing/destinations.rb
```diff
@@ -12,6 +12,26 @@
   acts_as_async_destroy('Routing::Destination')
   acts_as_async_update BatchUpdateForm::Destination
 
+  scoped_collection_action :asynce_schedule_rate_changes,
```
gigorok, line 15: typo

`async` instead of `asynce`

app/admin/routing/destinations.rb
```diff
@@ -12,6 +12,26 @@
...
+                           title: 'Schedule rate changes',
+                           class: 'scoped_collection_action_button ui',
+                           form: -> { Destination::ScheduleRateChangesForm.form_data },
+                           if: -> { authorized?(:batch_update, resource_klass) } do
+    attrs = params[:changes]&.permit!
```
gigorok, line 20: could we define specific params instead of accepting everything?

app/admin/routing/destinations.rb
```diff
@@ -12,6 +12,26 @@
...
+    if attrs.present?
+      form = Destination::ScheduleRateChangesForm.new(attrs)
+      form.ids_sql = scoped_collection_records.select(:id).to_sql
+
+      if form.save
+        flash[:notice] = 'Rate changes is scheduled'
+      else
+        flash[:error] = "Validation Error: #{form.errors.full_messages.to_sentence}"
+      end
+    end
```
gigorok, line 31: else flash error that params are required?

app/jobs/worker/schedule_rate_changes.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Worker
+  class ScheduleRateChanges < ApplicationJob
+    queue_as 'rates'
```
gigorok, line 5: do we have this queue?

spec/services/destination/schedule_rate_changes_spec.rb
```diff
@@ -0,0 +1,161 @@
...
+                               destination_id: destination.id
+                             )
+    end
+  end
+
+  context 'when destinations already have destinations next rates' do
+    let!(:destination_next_rates) do
+      [
+        *FactoryBot.create_list(:destination_next_rate, 3, destination: destinations[0]),
+        *FactoryBot.create(:destination_next_rate, destination: destinations[1])
```
gigorok, line 70: do we need the star here?

gigorok (commented):
also please attach some screenshots

app/forms/destination/schedule_rate_changes_form.rb
```diff
@@ -0,0 +1,49 @@
...
+    attribute :next_rate, :decimal
+    attribute :connect_fee, :decimal
+    attribute :ids_sql, :string
+
+    validates :initial_rate, :next_rate, :connect_fee, numericality: true
+    validates :initial_interval, :next_interval, numericality: { greater_than: 0 }
+    validates :ids_sql, :apply_time, presence: true
+    validate :validate_apply_time
+
+    def self.form_data
```
Skumring, line 18: In my opinion better name it `def self.form_inputs`

app/forms/destination/schedule_rate_changes_form.rb
```diff
@@ -0,0 +1,49 @@
...
+      {
+        apply_time: 'datepicker',
+        initial_interval: 'text',
+        initial_rate: 'text',
+        next_interval: 'text',
+        next_rate: 'text',
+        connect_fee: 'text'
```
Skumring, line 25: Also maybe all inputs except `apply_time` should be `:number`?

app/forms/destination/schedule_rate_changes_form.rb
```diff
@@ -0,0 +1,49 @@
...
+  class ScheduleRateChangesForm < ApplicationForm
+    attribute :apply_time, :date
+    attribute :initial_interval, :integer
+    attribute :initial_rate, :decimal
+    attribute :next_interval, :integer
+    attribute :next_rate, :decimal
+    attribute :connect_fee, :decimal
+    attribute :ids_sql, :string
+
+    validates :initial_rate, :next_rate, :connect_fee, numericality: true
```
Skumring, line 13: Can they be negative? I mean what happens if we enter '-1.0'?

Skumring (changes requested):

app/forms/destination/schedule_rate_changes_form.rb
BigG1947: It's looks like active_admin_scoped_collection_actions does not supported 'number' for inputs
https://github.com/activeadmin-plugins/active_admin_scoped_collection_actions/blob/master/vendor/assets/javascripts/lib/dialog_mass_fields_update.js
```js
    if (/^(datepicker|checkbox|text)$/.test(type)) {
      wrapper = 'input';
    } else if ($.isArray(type)) {
      [wrapper, elem, opts, type] = Array.from(['select', 'option', type, '']);
    } else {
      throw new Error(`Unsupported input type: {${name}: ${type}}`);
    }
```

app/forms/destination/schedule_rate_changes_form.rb
BigG1947: Those validations just copied from DestinationNextRate model. And already we allow to create such next_rates with negative initial rate, next rate and connect fee. But i agree what it's looks wrong. Maybe need add validation for DestinationNextRate.

app/assets/stylesheets/active_admin/components/active_admin_scoped_collection_actions.scss
```diff
@@ -1,6 +1,8 @@
 .ui-dialog.active_admin_dialog_mass_update_by_filter {
   width: auto!important;
   form ul li {
+    line-height: 30px;
```
BigG1947, line 4: *For fix batch actions forms

app/forms/destination/schedule_rate_changes_form.rb
gigorok: let's do this in the separated issue

PR #1472, opened 2024-05-15
schedule destinations rate changes improvement

BigG1947:
add option to create next rates with specifyс rate fields


app/services/destination/schedule_rate_changes.rb
```diff
@@ -20,27 +20,30 @@ def call
...
-            apply_time:,
-            destination_id:
-          }
-        end
-
-        Routing::DestinationNextRate.insert_all!(next_rate_attrs)
+        SqlCaller::Yeti.execute <<-SQL.squish
+          INSERT INTO class4.destination_next_rates (destination_id, initial_interval, initial_rate, next_interval, next_rate, connect_fee, apply_time)
+          SELECT id,
+                 #{initial_interval || 'initial_interval'},
```
dmitry-sinina, line 26: this is doesn't look ok. There is possibility for SQL injection. Can't we use ORM features there?

Skumring (approved):

app/services/destination/schedule_rate_changes.rb
dmitry-sinina: also may be it will be better to allow creation class4.destination_next_rates with null in rates/interval/fee  and not update such attributes during applying?

Skumring: we can use sanitize_sql, @BigG1947 take a look please

BigG1947: > also may be it will be better to allow creation class4.destination_next_rates with null in rates/interval/fee and not update such attributes during applying?

It's look like one more possible way. I will do this. But one little moment to clarify: 
If we create class4.destination_next_rates with null such next_rates will be displayed on index page with empty cells. In my opinion It may not be entirely clear how such records will affect for destination record. As suggestion we can decorate next_rates index page and show value that destination already has instead of 'EMPTY'. Or we can just keep it empty.
What do you think? @Skumring @gigorok @dmitry-sinina 

Thx

Skumring (approved):

PR #1487, opened 2024-06-17
fix performance issue with loading a lot of records on form inputs

BigG1947:


dmitry-sinina:
Same feature should be added to routing plans too(there is also numberlist selection)

PR #1508, opened 2024-07-17
allow to ContactEmailSender to create email log even when there is SMTP setting

### LINKED ISSUE #1395: Jobs::AccountBalanceNotify exception if no smtp connections available
```
irb(main):001:0> p=Jobs::AccountBalanceNotify.new(0)
=> #<Jobs::AccountBalanceNotify:0x00007f3ff709bbc0 @options=0>
irb(main):002:0> p.execute
/opt/yeti-web/app/domain/contact_email_sender.rb:47:in `create_email_log': undefined method `from_address' for nil:NilClass (NoMethodError)

      mail_from: contact.smtp_connection.from_address,
                                        ^^^^^^^^^^^^^
irb(main):003:0> 
irb(main):004:0> System::SmtpConnection.all
=> []

```

Ivanov-Anton:
### Description

when we perform `ContactEmailSender.batch_send_emails(contacts, subject: 'text', message: 'message')`

the Email log is not created due there is no any SMTP configs even global

To imptove this logic I suggest to remove `not null` constraint from `mail_from` column of `notifications.email_logs` table and do not fill in this column when we perform `batch_send_emails` method.
As result we will able to view Email Log (Logs->Email Log) and determinate that curently there is no any SMTP config and it should be created for specific contact or Global one.

### Additional links

closes #1395

### DIFF
--- a/spec/domain/contact_email_sender_spec.rb
+++ b/spec/domain/contact_email_sender_spec.rb
@@ -23,7 +23,6 @@ RSpec.describe ContactEmailSender do
         expect { subject }.to change { Log::EmailLog.count }.by(1)
         email_log = Log::EmailLog.last!
         expect(email_log).to have_attributes(expected_email_log_attrs)
-        expect(Worker::SendEmailLogJob).to have_been_enqueued.with(email_log.id)
       end
 
       it 'enqueues Worker::SendEmailLogJob' do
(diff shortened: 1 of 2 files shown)


spec/domain/contact_email_sender_spec.rb
```diff
@@ -23,7 +23,6 @@
         expect { subject }.to change { Log::EmailLog.count }.by(1)
         email_log = Log::EmailLog.last!
         expect(email_log).to have_attributes(expected_email_log_attrs)
-        expect(Worker::SendEmailLogJob).to have_been_enqueued.with(email_log.id)
```
Ivanov-Anton, line 26: removed because at the 31 line we have already perform this test

PR #1522, opened 2024-07-25
allow to create network_prefixes with the same prefix

### LINKED ISSUE #1521: System->Network prefixes, check number length
Multiple objects with same prefix but different number lengths should be allowed.

Ivanov-Anton:
### Description

allow to create network_prefixes with the same prefix but diff length of number

closes #1521

### DIFF
--- a/app/models/system/network_prefix.rb
+++ b/app/models/system/network_prefix.rb
@@ -14,7 +14,7 @@
 #
 # Indexes
 #
-#  network_prefixes_prefix_key        (prefix) UNIQUE
+#  network_prefixes_prefix_key        (prefix,number_min_length,number_max_length) UNIQUE
 #  network_prefixes_prefix_range_idx  (((prefix)::prefix_range)) USING gist
 #  network_prefixes_uuid_key          (uuid) UNIQUE
 #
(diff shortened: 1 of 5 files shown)


db/structure.sql
```diff
@@ -39847,10 +39839,10 @@ CREATE UNIQUE INDEX "index_sys.cdr_exports_on_uuid" ON sys.cdr_exports USING btr
 
 
 --
--- Name: network_prefixes_prefix_range_idx; Type: INDEX; Schema: sys; Owner: -
+-- Name: network_prefixes_prefix_with_scope_idx; Type: INDEX; Schema: sys; Owner: -
 --
 
-CREATE INDEX network_prefixes_prefix_range_idx ON sys.network_prefixes USING gist (((prefix)::public.prefix_range));
```
dmitry-sinina, line 39853: prefix_range index by prefix shouldn't be modified. We have only change UNIQUE  index.

app/models/system/network_prefix.rb
```diff
@@ -51,7 +50,7 @@ class System::NetworkPrefix < ApplicationRecord
               if: :number_max_length
             }
 
-  validates :prefix, uniqueness: { allow_blank: true }, presence: true
+  validates :prefix, uniqueness: { allow_blank: true, scope: [:number_min_length, :number_max_length] }, presence: true
```
Ivanov-Anton, line 53: I think it would be great to create the custom unique validation to mark the three fields as invalid.

And in admin form when submit an invalid form, the fields number_max_length number_min_length and prefix will be red.

dmitry-sinina: It is not important currently

app/models/system/network_prefix.rb
```diff
@@ -51,8 +51,9 @@ class System::NetworkPrefix < ApplicationRecord
               if: :number_max_length
             }
 
-  validates :prefix, uniqueness: { allow_blank: true }, presence: true
+  validates :prefix, presence: true
   validates :network, presence: true
+  validate :prefix_unique_validation, if: :prefix?, on: :create
```
dmitry-sinina, line 56: shouldn't we validate on update too?

PR #1527, opened 2024-08-09
config to control the behavior of the CallsMonitoring job

### LINKED ISSUE #1526: CallsMonitoring teardown rules
Currently CallsMonitoring terminating calls in case of disabled/deleted customer auth, origination gw, termination gw.
We have to allow flexible configuration for this. 

1. add configuration attributes in yeti_web.yml like teardown_on_disabled_customer_auth: false, teardown_on_disabled_term_gw: false
2. Current behavior should be default when no configs present

Ivanov-Anton:
## Description

add config to control the behavior of the CallsMonitoring job

introduces new configuration options in yeti_web.yml to control
the behavior of the CallsMonitoring job when dealing with
disabled customer authentication and both origination/termination gateways:

```
- calls_monitoring.teardown_on_disabled_customer_auth
- calls_monitoring.teardown_on_disabled_term_gw
- calls_monitoring.teardown_on_disabled_orig_gw
```

The CallsMonitoring class has been updated to use these
configurations in the relevant methods. If the configs
are not present, the current flow should not be changed.

## Additional links

closes #1526

### DIFF
--- a/app/jobs/jobs/calls_monitoring.rb
+++ b/app/jobs/jobs/calls_monitoring.rb
@@ -165,8 +165,12 @@ module Jobs
         detect_vendors_calls_to_reject
       end
 
-      log_time('detect_gateway_calls_to_reject') do
-        detect_gateway_calls_to_reject
+      log_time('detect_orig_gateway_calls_to_reject') do
+        detect_orig_gateway_calls_to_reject
+      end
+
+      log_time('detect_term_gateway_calls_to_reject') do
+        detect_term_gateway_calls_to_reject
       end
 
       log_time('detect_random_calls_to_reject') do
@@ -187,8 +191,6 @@ module Jobs
         terminate_calls!
       end
     end
-
-    private
 
     # random_disconnect_enable        | f
     # random_disconnect_length        | 7000
@@ -273,9 +275,18 @@ module Jobs
       end
     end
 
+    def teardown_enabled?(config_key)
+      value = YetiConfig.calls_monitoring.send(config_key)
+      return true if value.nil? # Default behavior
+
+      value
+    end
+
     # drop calls where `customer_auth_id`
     # is linked to `CustomersAuth#reject_calls = true`
     def detect_customers_auth_calls_to_reject
+      return unless teardown_enabled?(:teardown_on_disabled_customer_auth)
+
       flatten_calls.each do |call|
         customers_auth_id = call[:customer_auth_id]
         customers_auth = active_customers_auths_reject_calls[customers_auth_id]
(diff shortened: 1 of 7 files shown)


app/jobs/jobs/calls_monitoring.rb
```diff
@@ -188,8 +188,6 @@ def before_finish
       end
     end
 
-    private
```
Ivanov-Anton, line 191: removed because duplicated.

Ivanov-Anton:
@dmitry-sinina this PR is ready to code review

Ivanov-Anton:
@dmitry-sinina this PR is ready to code review

app/jobs/jobs/calls_monitoring.rb
```diff
@@ -289,13 +296,28 @@ def detect_customers_auth_calls_to_reject
...
-        if disabled_orig_gw_active_calls.key?(call[:orig_gw_id]) || disabled_term_gw_active_calls.key?(call[:term_gw_id])
-          local_tag = call[:local_tag]
-          @terminate_calls[local_tag] = call
+        if disabled_orig_gw_active_calls.key?(call[:orig_gw_id])
+          terminate_calls_for_disabled_orig_gw(call)
+        elsif disabled_term_gw_active_calls.key?(call[:term_gw_id])
+          terminate_calls_for_disabled_term_gw(call)
         end
       end
     end
```
senid231, line 305: I think better would be to split `detect_gateway_calls_to_reject` into 2 methods: `detect_orig_gateway_calls_to_reject` and `detect_term_gateway_calls_to_reject` and use same approach that was used in `detect_customers_auth_calls_to_reject`

**P.S.** also please refactor adding calls to `@terminate_calls`, they all should use `terminate_calls` method

```ruby
    def detect_orig_gateway_calls_to_reject
      return unless teardown_enabled?(:teardown_on_disabled_orig_gw)

      calls_to_terminate = flatten_calls.select do |call|
        disabled_orig_gw_active_calls.key?(call[:orig_gw_id])
      end
      terminate_calls(calls_to_terminate)
    end

    def detect_term_gateway_calls_to_reject
      return unless teardown_enabled?(:teardown_on_disabled_term_gw)

      # same
    end

    def terminate_calls(calls)
      @terminate_calls.merge!(calls.index_by { |c| c[:local_tag] })
    end
```

senid231 (changes requested):

Ivanov-Anton:
@senid231 appliyed your last suggestion.

dmitry-sinina:
new config variables should be described there:

https://github.com/yeti-switch/yeti-web/blob/master/config/initializers/_config.rb#L25

Ivanov-Anton:
just appliyed your last suggestion

PR #1551, opened 2024-09-09
1526, do not require teardown configs, they should be optional

Ivanov-Anton:
### Description

do not require 
```
teardown_on_disabled_customer_auth
teardown_on_disabled_term_gw
teardown_on_disabled_orig_gw
```
configs, they should be optional

## Additional links

related to #1526


Ivanov-Anton:
related to #1552

PR #1682, opened 2025-01-15
add ability to sort destinations by network type by customer order

BigG1947:


db/migrate/20250115133012_add_sorting_priority_to_network_types.rb
```diff
@@ -0,0 +1,40 @@
...
+UPDATE network_types
+SET sorting_priority = network_type_priorities.priority
+FROM network_type_priorities
+WHERE network_types.name = network_type_priorities.name
+    SQL
+
+    execute <<~SQL
+UPDATE network_types
+SET sorting_priority = 999
+WHERE sorting_priority IS NULL
```
gigorok, line 33: don't think that update old data is needed, cause names can be different in different installations. default value will be ok

db/migrate/20250115133012_add_sorting_priority_to_network_types.rb
```diff
@@ -0,0 +1,40 @@
+class AddSortingPriorityToNetworkTypes < ActiveRecord::Migration[7.0]
+  def up
+    add_column :network_types, :sorting_priority, :integer, default: 999
```
dmitry-sinina, line 3: smallint please

PR #1706, opened 2025-03-06
ref-1704, destination_next_rates idx by destination_id

gigorok:
in order to speed up deleting outdated destination_rates, we have to add missed index by destination_id


gigorok:
duplicate here - https://github.com/yeti-switch/yeti-web/pull/1705

PR #1725, opened 2025-04-07
Admin UI: disallow deletion timezones records from the Admin

Ivanov-Anton:


dmitry-sinina:
can't we just remove this admin resource  
app/admin/system/time_zones.rb
 ?

Issue #1728, opened 2025-04-14
CDR view policy attributes


dmitry-sinina:
```
      Cdr/Cdr:
        read: true
        pcap: true
        recording: true
        dst_number: true
```

*pcap* - should be true to allow pcap trace download
*recording* - should be true to allow recording download
*dst_number* - should be true to show dst_prefix_in,dst_prefix_routing, dst_prefix_out, legb_ruri. when dst_number=false(by default) then last 3 digits of numbers should be masked

PR #1741, opened 2025-05-06
#1728, Refactor CdrDecorator to use constant for phone number masking

Ivanov-Anton:
- Introduced MASK_PHONE_NUMBER_REGEXP in Cdr model for consistent phone number masking.
- Updated decorated_legb_ruri method in CdrDecorator to utilize the new constant.

### DIFF
--- a/app/policies/cdr/cdr_policy.rb
+++ b/app/policies/cdr/cdr_policy.rb
@@ -4,8 +4,11 @@ module Cdr
   class CdrPolicy < ::RolePolicy
     section 'Cdr/Cdr'
     self.allowed_actions += %i[allow_full_dst_number recording pcap]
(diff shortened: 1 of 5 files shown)


app/policies/cdr/cdr_policy.rb
```diff
@@ -4,7 +4,7 @@ module Cdr
   class CdrPolicy < ::RolePolicy
     section 'Cdr/Cdr'
     self.allowed_actions += %i[allow_full_dst_number recording pcap]
-    alias_rule :dump?, :routing_simulation?, to: :perform?
+    alias_rule :routing_simulation?, to: :perform?
```
dmitry-sinina, line 7: I think **routing_simulation** should depends on access to **RoutingSimulation** page because this button just redirect user to other page

PR #1757, opened 2025-05-30
write logs to stdout instead of file

BigG1947:


config/environments/production.rb
```diff
@@ -48,16 +48,6 @@
...
-  # Use a different logger for distributed setups.
-  if ENV['RAILS_LOG_TO_STDOUT'].present?
-    STDOUT.sync = true
-    logger = ActiveSupport::Logger.new(STDOUT)
-    logger.formatter = config.log_formatter
-    config.logger = ActiveSupport::TaggedLogging.new(logger)
-  else
-    config.logger = ActiveSupport::TaggedLogging.new(Logger::Syslog.new('yeti-web', Syslog::LOG_LOCAL7))
-  end
-
```
BigG1947, line 60: but in my opinion we can keep manage about log way and do something like:

```ruby
  if ENV['RAILS_LOG_TO_STDOUT'].present?
  $stdout.sync = true
  config.rails_semantic_logger.add_file_appender = false
  config.semantic_logger.add_appender(
    io: $stdout,
    formatter: :default
  )
  end
```

PR #1761, opened 2025-06-11
filter cdrs by customer external ID

gigorok:
Screenshot 2025-06-13 at 15 36 31


gigorok:
the filter should be added as last

PR #1811, opened 2025-09-16
use Tempfile class during working with temporary files

BigG1947:


app/services/send_report/base.rb
```diff
@@ -99,11 +96,12 @@ def generate_csv_file(data_item)
           csv << line
         end
       end
-      file_name
     end
 
-    def csv_file_name
-      '/tmp/' + make_tmpname(["#{report.id}_#{report.class.to_s.parameterize}", '.csv'], nil)
+    def with_tmp_file
+      Tempfile.create(["#{report.id}_#{report.class.to_s.parameterize}", '.csv'], YetiConfig.tmpdir) do |tmp_file|
```
gigorok, line 102: In case when tmpdir is not configured, then `/tmp` should be used

PR #1821, opened 2025-09-26
Admin UI: Implement sorting by Network for Destination and Dialpeer pages

Skumring:
Screenshot 2025-09-25 at 17 47 59
Screenshot 2025-09-25 at 17 48 03
screenshot_2025-09-26-05-33-28 294
screenshot_2025-09-26-05-33-31 001


dmitry-sinina:
Could you test this sorting performance with 1-5M of Destinations?

Skumring:
> Could you test this sorting performance with 1-5M of Destinations?

I'll try but this test will be very approximately, because correct performance test is possible only on real data with a lot of associations, for example on Staging DB or something like that.

PR #1860, opened 2025-11-08
Add admin UI session expiration configuration

### LINKED ISSUE #1859: configurable cookie session expiration
### Description
Add configurable cookie session expiration

file `yeti_web.yml`

```yml
admin:
  session_lifetime: 120
```

Ivanov-Anton:
## Description

Introduce optional session expiration settings for the admin UI in the configuration files. Update `yeti_web.yml.ci`
Update `yeti_web.yml.distr` 

### example of new config
config/yeti_web.yml
```yml
admin_ui:
  session_lifetime: 60
```

## Additional links

closes #1859

### DIFF
--- a/config/initializers/config.rb
+++ b/config/initializers/config.rb
@@ -102,6 +102,10 @@ Config.setup do |setup_config|
       optional(:auto_approve).value(:bool)
     end
     optional(:default_ldap_roles).array(:string)
+
+    optional(:admin_ui).schema do
+      optional(:session_lifetime).maybe(:int?)
+    end
   end
 end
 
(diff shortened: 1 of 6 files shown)


Ivanov-Anton:
In this pull request, there’s one issue:
The new feature — which adds a property to the YAML config and handles sessions with expiration — is properly covered by integration tests. However, the case where the session_lifetime config is not set (the current default behavior) isn’t tested.

The reason is that I haven’t found a way to stub YetiConfig.admin_ui.session_lifetime. Using before { allow(YetiConfig).to receive(:admin_ui) } doesn’t work, because by the time the tests run, the Rails initialization process has already completed, and the session store is configured during that initialization stage.

PR #1863, opened 2025-11-11
[no issue] quieter, safer, and more readable test/CI output

Ivanov-Anton:
## Motivation

Occasionally, our test runs (bundle exec rspec some_spec.rb) print stray debug output instead of only the expected green dots. These messages come from temporary helpers like puts, p, pp, ap, warn, debugger/byebug, and binding.pry that were left behind during local debugging

Result: quieter, safer, and more readable test/CI output with near-zero ongoing cost.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -245,6 +245,10 @@ Lint/CircularArgumentReference:
   Enabled: true
 Lint/Debugger:
   Enabled: true
+  DebuggerMethods:
+    - p
+    - pp
+    - puts
 Lint/DeprecatedClassMethods:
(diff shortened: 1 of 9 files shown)


Ivanov-Anton:
> How to see debug messages after these changes?

You're 100% correct: after replacing puts with `Rails.logger.debug`, those messages no longer appear unless the Rails log level includes debug, and by default:

development → log level = debug → OK

production or test → log level = info or warn → debug messages are hidden

That's why you don’t see anything.

--------------

commit 100% backward-compatible because the `info` log level is visible in the development/production/test environments

PR #1878, opened 2025-12-11
Spelling fix

hcaldicott:
This is my first contribution to Yeti, and my first use of Ruby - I have never written a line of ruby before.

Would greatly appreciate it if one of the team could give this a once over - it is just a basic fix for a spelling mistake.

### DIFF
--- a/db/migrate/20251030125603_gw_group_spelling_fix.rb
+++ b/db/migrate/20251030125603_gw_group_spelling_fix.rb
@@ -1,26 +1,26 @@
 class GwGroupSpellingFix < ActiveRecord::Migration[7.2]
   def up
     execute %q{
-      update class4.codecs SET name = 'Priority/Weight balancing' 
+      update class4.gateway_group_balancing_modes SET name = 'Priority/Weight balancing' 
       where name = 'Priority/Weigth balancing';
 
-      update class4.codecs SET name = 'Priority/Weight balancing. Prefer gateways from same POP' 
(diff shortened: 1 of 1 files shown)


db/migrate/20251030125603_gw_group_spelling_fix.rb
```diff
@@ -0,0 +1,27 @@
+class GwGroupSpellingFix < ActiveRecord::Migration[7.2]
+  def up
+    execute %q{
+      update class4.codecs SET name = 'Priority/Weight balancing'
```
dmitry-sinina: Something very wrong there, it is not clear why you want to change CODEC name to "Priority/Weight balancing"

hcaldicott: Because it is spelt wrong originally, that's the point of this fix - 'Weigth' should be spelt 'Weight'. 

The idea was to correct the spelling in the front end interface, then ensure that the configuration in the database matches the newly corrected spelling. I can't fix the front-end without ensuring the database matches?

hcaldicott: I just realised I hadn't commited the correct version of the spelling fix database change. Corrected. I based this off another example change which was originally for the codecs table and didn't commit the final copy!

dmitry-sinina:
fixed in https://github.com/yeti-switch/yeti-web/pull/1912

PR #1980, opened 2026-04-10
add OIDC SSO authentication mode via activeadmin-oidc

Fivell:
- Add AdminUserOidcHandler concern (omniauthable + trackable + ip_allowable, password auth disabled via valid_password? returning false)
- File-gated mode selection: config/oidc.yml > config/ldap.yml > database
- AdminUser.external_auth? helper replaces ldap?-only guards in admin UI
- on_login hook filters IdP roles against AdminUser.available_roles, falls back to default_roles, denies login when no role matches
- Migration adds provider, uid, oidc_raw_info columns + partial unique index
- Login view conditionally hides password form in OIDC mode, uses form_tag + submit_tag for SSO button to match ActiveAdmin theme CSS
- Gem sourced from github: [activeadmin-plugins/activeadmin-oidc](https://github.com/activeadmin-plugins/activeadmin-oidc)


Fivell:
image

PR #1981, opened 2026-04-10
Preserve admin API authentication in OIDC mode

Ivanov-Anton:
## Description

When config/oidc.yml is present, AdminUser#authenticate is aliased to valid_password?, which now unconditionally returns false. Authentication::AdminAuth.authenticate! still calls admin_user.authenticate(@password) for POST /api/rest/admin/auth, so every admin API login starts failing as soon as OIDC is enabled, even for existing users with allowed IPs. That is a user-visible regression for any integration that relies on the documented admin JWT endpoint.


Fivell:
hi @Ivanov-Anton  thanks for pointing to it, i created regression branch with better tests and going to merge your code to this branch https://github.com/yeti-switch/yeti-web/pull/1982/commits

PR #2060, opened 2026-07-10
Speed up rspec CI: runtime-balanced groups + cached network-data seed + lighter filter matrix

senid231:
image

Three independent CI speedups for the rspec gate, all graceful: a missing cache or log falls back to today's behaviour, so there is no first-run regression and local dev is unaffected. Implements the plan in `docs/plans/rspec-parallel-test-distribution/plan.md`.

## Phase 1 — Balance parallel groups by measured runtime

The 8-way `parallel_test` matrix split by file **byte size** (the default when `--only-group` is set), which is uncorrelated with runtime and produced a ~4x compute imbalance across groups (worst ~812s vs best ~199s), setting the matrix critical path at ~15m51s.

- **Makefile**: pass `--group-by runtime --runtime-log $(RSPEC_RUNTIME_LOG) --allowed-missing 75`, gated by a `$(wildcard)` guard so a missing log falls back to filesize (safe first run / local dev). The file list still comes from scanning `spec/` on disk, so the log only supplies weights — it never skips or phantom-runs a file; unknown (new/renamed) files run at the average measured weight.
- **tests.yml**: restore the complete runtime log before `Run rspec`, upload each group's partial slice after, and a job that merges the 8 slices and caches the complete log for the next run. It runs on PRs too — caches are branch-scoped, so a PR reads its own measured log first and falls back to master's via `restore-keys` without poisoning the shared key. Gated on `!cancelled()` so a flaky node can't skip persistence; a `concurrency` group cancels superseded PR runs, closing the window where an overlapping run could publish a new log mid-matrix and desync the split.

## Phase 2 — Cache only the network-data seed across runs

Every DB-dependent job rebuilt the test DB from scratch. The dominant repeated cost is `custom_seeds[network_prefixes]`, which parses **~96 MB across 232 `db/network_data/*.yml` files** and bulk-inserts them — a fixed cost paid **11x per run** (rspec×8, oidc, database_consistency, annotations).

Rather than cache the whole DB (a dump keyed on schema/seeds is invalidated by every migration → 96 MB re-parsed everywhere), cache **only that step's result — the 4 `sys` network tables** — and rebuild the rest fresh each run, so the key is independent of `db/structure.sql` and `db/seeds` and survives migrations and base-seed edits.

- **Makefile** (`prepare-test-db` CI branch): always `db:create` + `db:schema:load` + `db:seed` (fast; `db:seed` also re-creates the current-month CDR partitions, so no separate partition task is needed). On a cache **miss**, run `custom_seeds[network_prefixes]` then `pg_dump --data-only` the 4 `sys` network tables (`network_types`, `networks`, `countries`, `network_prefixes`); on a **hit**, `DELETE` those tables (child→parent) and `pg_restore --data-only --disable-triggers --exit-on-error` the dump, resetting sequences and skipping the YAML load. `$(wildcard)`-guarded so local/first-run falls back to the full build.
- **.github/actions/restore-test-db-dump**: one composite action defines the `actions/cache` step, keyed **only on the network-data source files** (not `structure.sql`/`db/seeds`), reused by all 4 DB-prep jobs.
- **dev/db_test_prepare.sh**: the same network-data cache locally, keyed by a content hash embedded in the dump filename; degrades to a plain YAML load when the pg client is missing or older than the server.

The CI build image ships `postgresql-client-18` (`debian/control` Build-Depends), matching the PG18 service, so `pg_dump`/`pg_restore` are version-compatible (and a `--data-only` COPY dump is version-portable anyway).

## Phase 3a — Cut wasted rspec compute (node-count-independent)

Reduces *total* compute (the 2980s budget) and the whole-file floor, so it pays off at any node count.

- **Collapse the six `jsonapi_filters_by_*_field` shared examples** into one `aggregate_failures` example each: create the covering records **once** and share them across every operator instead of re-creating a record pair and re-issuing a request per operator context. Per `it_behaves_like` this drops record inserts 32→2 (string 3) and examples 16→1, compounding across the ~28 heavy controller/request specs that use them (cdrs, gateways, auth_logs, incoming_cdrs, customers_auths, dialpeers, destinations, …). New `assert_filter` + `perform_ransack_filter_request` helpers in `ransack_filter_setup` re-issue the index request per operator and read `response_data` directly; controller vs request shape is picked by `respond_to?(:json_api_request_path)`. The legacy `filter_key`/`filter_value` merge is preserved but guarded so collapsed examples fall through to the spec's base query. Correctness coverage is unchanged; on failure `aggregate_failures` still reports the offending operator.
- **Filter the network-prefix and network CSV exports** (`export_network_prefix_spec`, `export_network_spec`) by a dedicated parent (`country_id_eq` / `type_id_eq`) so they render a handful of rows instead of serializing the whole seeded `sys.network_prefixes` (~486k rows) and `sys.networks` tables through per-row name procs just to assert a couple of rows. Other export siblings render only test-created rows (empty tables), so they are left unchanged.

## Notes

- **Runtime-log reporting**: the persist job prints the 50 slowest spec files to the job log and run summary so heavy files can be spotted for future splitting (Phase 3b, deferred). It runs after the cache save so a reporting glitch can't block log persistence.
- **Review fix (folded into commit 1)**: the merge job originally saved the runtime log at repo-root `parallel_runtime_rspec_full.log` while the rspec job restored `log/parallel_runtime_rspec_full.log`. Since `actions/cache` derives its cache version from the `path` input, the mismatch meant the cache would never resolve. Both sides now use `log/parallel_runtime_rspec_full.log`.
- **Makefile `$(info)` syntax fix**: a literal `(slow)` in an `$(info:msg=...)` message closed make's variable reference early, spilling a stray `)` outside `printf`'s quotes and aborting `prepare-test-db` on every cache-miss run. Parens dropped; both cold (YAML-load) and warm (`pg_restore`) DB-prep branches verified green.

Phases 3b (file splits) and 4 (node-count re-tune) are deferred — not needed at 8 nodes.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,7 +2,15 @@ name: Tests
 on:
   pull_request:
   push:
-    branches: [ master ]
+    branches: [master]
+
+# Cancel a superseded run when new commits are pushed to the same PR. Besides saving
+# CI minutes, this closes the narrow window where two overlapping runs of the same
+# branch could publish a new runtime-log cache mid-matrix and desync the 8-way split.
+# Push runs on master are never cancelled, so every master run persists its log.
+concurrency:
+  group: tests-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
   gems-caching:
@@ -33,7 +41,7 @@ jobs:
           key: gems-deb13-${{runner.os}}-${{hashFiles('Gemfile.lock')}}-${{hashFiles('.ruby-version')}}
 
   lint:
-    name: Rubocop and bundle audit
+    name: Rubocop
     runs-on: ubuntu-latest
     container: ghcr.io/yeti-switch/yeti-web/build-image:trixie
     needs: gems-caching
@@ -47,8 +55,26 @@ jobs:
             .bundle
             /opt/yeti-web/vendor/rbenv
           key: gems-deb13-${{runner.os}}-${{hashFiles('Gemfile.lock')}}-${{hashFiles('.ruby-version')}}
-      - name: Rubocop and bundle audit
+      - name: Rubocop
         run: make lint
+
+  audit:
+    name: Bundle audit
+    runs-on: ubuntu-latest
+    container: ghcr.io/yeti-switch/yeti-web/build-image:trixie
+    needs: gems-caching
+    steps:
+      - uses: actions/checkout@v6
+      - name: use cache
+        uses: actions/cache@v5
+        with:
+          path: |
+            vendor
+            .bundle
+            /opt/yeti-web/vendor/rbenv
+          key: gems-deb13-${{runner.os}}-${{hashFiles('Gemfile.lock')}}-${{hashFiles('.ruby-version')}}
+      - name: Bundle audit
+        run: make audit
 
   database_consistency:
     name: DB consistency check
@@ -76,6 +102,7 @@ jobs:
           YETI_DB_PORT: 5432
           CDR_DB_HOST: db
           CDR_DB_PORT: 5432
+          SEED_WORKERS: 4
 
       - name: Save artifacts
         if: always()
@@ -86,7 +113,6 @@ jobs:
           include-hidden-files: true
           path: |
             log/
-
 
   annotations:
     name: Model annotations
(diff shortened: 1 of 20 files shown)


Ivanov-Anton:
related Pull request that changes the grouping strategy into `runtime` mode

https://github.com/yeti-switch/yeti-web/pull/1922

PR #2061, opened 2026-07-10
aa4

dmitry-sinina:


Ivanov-Anton:
Agree with you. IT would be great to update ActiveAdmin to 4 version.

PR #2072, opened 2026-07-25
YETI-2071: Fix schema search path formatting

### LINKED ISSUE #2071: Prevent unnecessary structure.sql changes after migrations
Description:

```markdown
## Problem

Running database migrations causes an unrelated formatting change in `db/structure.sql`:

```diff
-SET search_path TO gui, public, switch, billing, class4, runtime_stats, sys, logs, data_import;
+SET search_path TO gui, public, switch, billing, class4, runtime_stats, sys, logs, data_import
+;
```

## Cause

The multiline `schema_search_path` values use YAML `>`, which preserves a trailing newline. Rails includes that newline when generating `structure.sql`.

## Expected behavior

Running migrations without schema changes should not produce unrelated changes in `db/structure.sql`.

## Proposed solution

Use YAML `>-` in all database configuration templates to strip the trailing newline and produce consistent output across Debian, Ubuntu, macOS Intel, and Apple Silicon.
```

Ivanov-Anton:
## Summary

Strip the trailing newline from `schema_search_path` in database configuration templates.

This prevents Rails from producing an unnecessary line break in `db/structure.sql` after migrations.

The output is now consistent across Linux and macOS environments.

## Related links

closes #2071

### DIFF
--- a/config/database.build.yml
+++ b/config/database.build.yml
@@ -7,7 +7,7 @@ production:
     username: postgres
     password:
     host: <%= (ENV['YETI_DB_HOST'] || 'localhost') %>
-    schema_search_path: >
+    schema_search_path: >-
       gui, public, switch,
       billing, class4, runtime_stats,
       sys, logs, data_import
--- a/config/database.yml.development
+++ b/config/database.yml.development
@@ -7,7 +7,7 @@ development:
     username: postgres
     password:
     host: 127.0.0.1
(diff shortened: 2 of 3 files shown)


dmitry-sinina:
lets use single line values, same as we doing in our documentation config examples: https://yeti-switch.org/docs/installation/installation-1.14/web.html#yeti-web-configuration

